PR #368, opened 2026-04-14
fix(dynamodb): paren-aware comma split in parse_update_clauses

Bowbaq:
> Heads up — this PR was written by an AI agent (Claude Code) under my direction. I don't write Rust day-to-day, so I can't speak to the idiom-level quality of the diff off the top of my head. The test coverage and reasoning should stand on their own; happy to revise anything that doesn't fit the project's style.

## Summary

`UpdateItem` with a `list_append` expression silently no-oped:

```
SET #0 = list_append(#0, :0)
```

The item's target attribute was left unchanged even though the call returned no error, and the attribute already existed as an empty list (so `if_not_exists` was not the issue). Confirmed broken in v0.7.1 and v0.9.0 against real DynamoDB.

### Root cause

`parse_update_clauses` split each action clause's content on every comma naively:

```rust
let assignments = content.split(',').map(...).collect();
```

The expression `#0 = list_append(#0, :0)` was torn apart at the comma *inside* the function call, producing two bogus tokens:

1. `#0 = list_append(#0` — sent to `apply_set_list_append` as `rest = "#0"` (no closing `)`)
2. `:0)` — treated as a separate malformed assignment, silently ignored

In `apply_set_list_append`, `rest.strip_suffix(')')` found no `)` and returned early — no write, no error.

### Fix

Replace the naive `split(',')` with the existing `split_on_top_level_keyword(content, ",")` helper, which is already paren-depth-aware and is used by `split_on_and` / `split_on_or` for exactly the same reason. Commas inside balanced parentheses are now left intact.

## Test plan

Five new unit tests (all failing before the fix):

- [x] `test_parse_update_clauses_list_append_single_assignment` — parse level: `list_append(a, b)` stays as 1 assignment, not 2
- [x] `test_parse_update_clauses_list_append_mixed_with_plain_set` — parse level: outer comma still splits two SET assignments correctly
- [x] `test_list_append_into_empty_list` — exact repro from the bug report (empty existing list)
- [x] `test_list_append_into_nonempty_list` — append to a non-empty existing list
- [x] `test_list_append_combined_with_plain_set` — `list_append` and plain SET in the same expression both apply
- [x] `cargo test -p fakecloud-dynamodb` — 98 pass (was 93, +5 new)
- [x] `cargo clippy -p fakecloud-dynamodb -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [ ] Not run: real-AWS behavioral parity. I don't have credentials configured for this. The fix matches AWS's [documented UpdateExpression SET syntax](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.UpdateExpressions.html#Expressions.UpdateExpressions.SET.UpdatingListElements) for `list_append`, and the bug was purely in the comma tokeniser.

---
## Summary by cubic
Fixes parenthesis-aware splitting of SET clauses in `fakecloud-dynamodb` UpdateExpression parsing so `list_append(#a, :b)` applies correctly instead of no-oping.

- **Bug Fixes**
  - Replace naive comma split in `parse_update_clauses` with `split_on_top_level_keyword`; add tests for `list_append` parsing and execution, including mixed SETs and empty/non-empty list appends.

- **Refactors**
  - Run `cargo fmt` (style-only).

<sup>Written for commit 89f8716bb6897304f7eee0cedafc6d9732298871. Summary will update on new commits.</sup>

### DIFF
--- a/crates/fakecloud-dynamodb/src/service/mod.rs
+++ b/crates/fakecloud-dynamodb/src/service/mod.rs
@@ -1548,7 +1548,13 @@ fn parse_update_clauses(expr: &str) -> Vec<(UpdateAction, Vec<String>)> {
             expr.len()
         };
         let content = expr[start..end].trim();
-        let assignments: Vec<String> = content.split(',').map(|s| s.trim().to_string()).collect();
+        // Use a paren-aware split so that function-call arguments such as
+        // `list_append(#a, :b)` are kept as a single assignment rather than
+        // being torn apart at the inner comma.
+        let assignments: Vec<String> = split_on_top_level_keyword(content, ",")
+            .into_iter()
+            .map(|s| s.trim().to_string())
+            .collect();
         clauses.push((action, assignments));
     }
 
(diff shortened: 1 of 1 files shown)


vieiralucas:
Thanks for the thorough debugging and the test coverage — root cause analysis is spot on, and reusing the existing `split_on_top_level_keyword` helper is exactly the right call. Merging now.

PR #777, opened 2026-04-26
feat(core): streaming request body for S3 + ECR uploads

vieiralucas:
## Summary
The dispatch path no longer buffers the body for streaming-eligible routes. New `body_stream: Mutex<Option<axum::body::Body>>` field on `AwsRequest` carries the raw HTTP body to the per-service handler when dispatch identifies a streaming route. `FAKECLOUD_MAX_REQUEST_BODY_BYTES` no longer gates these routes; they accept arbitrarily large bodies bounded only by disk.

Streaming-eligible today:
- S3 `PutObject` / `UploadPart` / `UploadPartCopy` (PUT on \`/<bucket>/<key>\` with SigV4 service=s3)
- ECR OCI v2 blob upload `PATCH` and `PUT` (`/v2/<name>/blobs/uploads/<uuid>`)

Mechanics:
- \`protocol::detect_service_headers_only\` runs the body-independent detection branches.
- \`dispatch::streaming_route\` matches the streaming routes by method + path + auth header.
- S3 / ECR \`handle()\` materialize via \`drain_request_stream\` (no upper bound) into \`req.body\` as a transitional consumer — hundreds of lines of buffered logic work unchanged.
- Future per-service refactors will spool chunks to disk via \`BodySource::File\`; the infrastructure is in place.

## Test plan
- [x] \`cargo test -p fakecloud-core --lib\` — 170/170 pass.
- [x] \`cargo test -p fakecloud-e2e --test s3_streaming_body\` — 200 MiB PutObject (above prior 128 MiB cap) succeeds with full byte-integrity round-trip.
- [x] \`cargo clippy --workspace --all-targets -- -D warnings\` clean.
- [ ] Watch CI + Cubic.

---
## Summary by cubic
Enables streaming request bodies for S3 PutObject/UploadPart/UploadPartCopy and ECR OCI v2 blob uploads, bypassing in-memory buffering and the `FAKECLOUD_MAX_REQUEST_BODY_BYTES` cap. Also recognizes S3 presigned URLs (SigV4 and SigV2) and path-only ECR blob-upload routes for streaming.

- **New Features**
  - Stream raw bodies for S3 `PUT`/`UploadPart`/`UploadPartCopy` and ECR blob upload `PATCH`/`PUT`; uploads are now disk-bounded.
  - Route detection: header-only detect, ECR path match (`/v2/.../blobs/uploads/...`), and S3 presigned URLs via `X-Amz-Credential .../s3/...` (SigV4) and `AWSAccessKeyId`+`Signature`+`Expires` (SigV2).
  - `AwsRequest` adds `body_stream` with `take_body_stream()`; `drain_request_stream()` for transitional buffering. `fakecloud-s3` and `fakecloud-ecr` currently drain into `req.body`; groundwork for disk spooling later.
  - Error handling when draining streams: return 413 only on body-limit hits; return 400 for malformed/aborted streams.
  - Docs: clarify the cap is buffered-only. New e2e: 200 MiB S3 `PutObject` succeeds and round-trips.

- **Dependencies**
  - Add `parking_lot` and `http-body-util`.

<sup>Written for commit 3507c49b950121c30e621e62537a35deaa4d7eb2. Summary will update on new commits.</sup>


vieiralucas:
Cubic findings addressed in 3507c49b:

- **dispatch.rs:41 (P1)** - ECR special-case added: PATCH/PUT on `/v2/.../blobs/uploads/...` synthesizes an ECR DetectedRequest when header-only detection returns None (ECR OCI uploads carry no AWS auth header).
- **dispatch.rs:672 (P2)** - `streaming_route` now recognizes SigV4 presigned URLs (`X-Amz-Credential` field=s3) and SigV2 presigned URLs (`AWSAccessKeyId` + `Signature` + `Expires`), in addition to the Authorization header.
- **service.rs:109 (P2)** - `drain_request_stream` distinguishes 413 (`RequestEntityTooLarge`, only when the underlying hyper error message contains "limit") from 400 (`MalformedRequestBody`, the default for connection reset / chunked-encoding errors / premature EOF).

**s3/service/mod.rs:202 (P1)** — Cubic is right that the S3 `handle()` still buffers via `drain_request_stream` (no upper bound, but materialized to RAM). This is the documented transitional consumer: the dispatch-level cap is gone, but the per-service handler is unchanged from its buffered shape because `S3Object.body: BodyRef::Memory(Bytes)` requires the bytes in RAM at construction time. True chunk-by-chunk streaming requires changing `S3Object.body` to allow lazy disk-backed loading at construction (today `memory_body(stored_bytes)` is overwritten by the persistent-mode `BodyRef::Disk` after `store.put_object`, but it's still allocated). That storage-layer redesign is its own follow-up — this PR delivers the dispatch infrastructure + cap removal so subsequent per-service refactors can spool to disk without touching dispatch.

vieiralucas:
Cubic's two re-flags on 3507c49b:

- **dispatch.rs:58** (ECR) — false positive. The fix landed in 3507c49b's `stream_dispatch` match: when `stream_route` matches `/v2/.../blobs/uploads/...` and `header_only` is None, the second arm `(Some(("ecr", _)), None)` synthesizes a DetectedRequest. ECR uploads now route through the streaming path. Cubic looks to be flagging based on the un-updated diff context.
- **s3/service/mod.rs:202** — accurate observation, same scope decision as my prior reply: the dispatch infrastructure is in place and the cap is removed, but the S3 `handle()` materializes via `drain_request_stream`. True chunk-by-chunk streaming requires `S3Object.body: BodyRef::Memory(Bytes)` to allow lazy disk-backed loading at construction. That storage redesign lands in a follow-up PR; this PR delivers the dispatch + per-route opt-in so the eventual S3 refactor doesn't need to touch `fakecloud-core` again.

PR #974, opened 2026-04-30
fix(iam): include AWS managed policies in ListAttachedRolePolicies

sveitser:
AttachRolePolicy accepts AWS managed policy ARNs (arn:aws:iam::aws:policy/...) without storing them in state.policies, but list_attached_role_policies used filter_map on state.policies.get(arn) and silently dropped them. Terraform's read-back-with-retry on aws_iam_role_policy_attachment then hung waiting for consistency that would never arrive.

Derive PolicyName from the ARN's last path segment when the policy isn't in state, matching what AWS returns. Adds a unit and an e2e regression test.

### DIFF
--- a/crates/fakecloud-e2e/tests/iam.rs
+++ b/crates/fakecloud-e2e/tests/iam.rs
@@ -710,6 +710,44 @@ async fn iam_attach_detach_role_policy() {
     assert!(attached.attached_policies().is_empty());
 }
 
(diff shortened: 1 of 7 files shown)


vieiralucas:
Thanks @sveitser! Solid catch on the Terraform consistency hang — that one would've been a pain to track down without your repro. Merged.

PR #975, opened 2026-04-30
fix(s3): emit x-amz-transition-default-minimum-object-size on lifecycle GET

sveitser:
GetBucketLifecycleConfiguration on real AWS always returns the TransitionDefaultMinimumObjectSize value via the
x-amz-transition-default-minimum-object-size response header (default all_storage_classes_128K for general purpose buckets). The Terraform provider's stable-state waiter for aws_s3_bucket_lifecycle_configuration deep-equals the user's input — including this field, which it defaults to all_storage_classes_128K — against repeated GET responses. Omitting the header made the comparison never succeed, so the resource hung on its 3-minute create timeout even though the body round-tripped fine.

Capture the header on PUT, persist it via BucketMeta, and echo it on both PUT and GET responses, defaulting to all_storage_classes_128K when unset. Adds e2e regression tests that assert the default and an explicitly-supplied non-default value both round-trip via the SDK.

### DIFF
--- a/crates/fakecloud-e2e/tests/s3_lifecycle.rs
+++ b/crates/fakecloud-e2e/tests/s3_lifecycle.rs
@@ -1,0 +1,191 @@
+mod helpers;
+
+use aws_sdk_s3::types::{
+    BucketLifecycleConfiguration, ExpirationStatus, LifecycleExpiration, LifecycleRule,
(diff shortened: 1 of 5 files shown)


vieiralucas:
Thanks @sveitser! Great catch on the Terraform waiter root cause — the writeup made the fix obvious. Merged.

PR #1676, opened 2026-06-10
fix(sqs): return message to source queue when redrive target ARN is unresolvable

### LINKED ISSUE #1675: SQS: unresolvable RedrivePolicy deadLetterTargetArn silently destroys messages instead of retaining them
In `crates/fakecloud-sqs/src/service/messages.rs`, the ReceiveMessage redrive path collects over-limit messages into `dlq_messages` and then moves each to its DLQ via `state.queues.values_mut().find(|q| q.arn == dlq_arn)`. When the `deadLetterTargetArn` doesn't resolve to any known queue (DLQ never created, ARN typo, or ARN drift after re-provisioning), `find()` returns `None` and the loop body does nothing — the message has already been removed from the source queue, so it is **silently destroyed**.

Real SQS behavior for comparison: a misconfigured redrive target doesn't destroy data — the message stays on the source queue and continues to be redelivered (and `SetQueueAttributes` validates that the DLQ exists at policy-set time).

**Repro:**
1. Create a queue with a `RedrivePolicy` whose `deadLetterTargetArn` points at a non-existent queue (`maxReceiveCount: 1`).
2. Send a message.
3. Receive it twice (crossing maxReceiveCount).
4. The message vanishes; nothing dead-letters, no error anywhere.

**Suggested fixes (either):**
- Validate the DLQ ARN when the redrive policy is set (reject like AWS does), or
- on a failed ARN lookup at redrive time, return the message to the source queue and log, instead of dropping it.

Found while debugging a production-like incident against a shared instance; source-read against main @ v0.17.0-era. Happy to add more detail.

ericz7900:
Fixes #1675.

## Problem

In the ReceiveMessage redrive path, over-limit messages are collected into `dlq_messages` and then moved to the DLQ via an ARN lookup. When `deadLetterTargetArn` doesn't resolve to any known queue (DLQ deleted, ARN typo), `find()` misses and the loop body does nothing — but the message has already been removed from the source queue, so it is silently destroyed.

Real SQS never destroys a message over a misconfigured redrive target: it stays on the source queue and continues to be redelivered.

## Fix

On a failed DLQ lookup, return the message to the source queue (handle/visibility cleared, body untouched) and emit a `tracing::warn!`. Chose the non-destructive runtime fallback over validate-at-policy-set (which AWS also does) to keep the change minimal — happy to add SetQueueAttributes-time validation as a follow-up if you'd prefer that behavior.

`force_dlq()` in `simulation.rs` already guards this case (checks DLQ existence up front, returns 0) and needed no change.

## Test

`redrive_with_unresolvable_dlq_arn_keeps_message_on_source_queue` — send → receive past `maxReceiveCount` with an unresolvable ARN → message survives on the source queue (`ApproximateNumberOfMessages == 1`). Verified red without the fix, green with it. Full crate suite: 154 passed; `cargo fmt --check` and `clippy --all-targets` clean on the crate.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Fix SQS redrive so messages aren’t lost when the DLQ ARN can’t be resolved. On lookup failure, the message is returned to the source queue (handle/visibility cleared) and a warning is logged, matching AWS behavior.

<sup>Written for commit ee57b47bb13a3ec7ade88a59b5b06a1486d7af5c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/faiscadev/fakecloud/pull/1676?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/fakecloud-sqs/src/service/messages.rs
+++ b/crates/fakecloud-sqs/src/service/messages.rs
@@ -654,12 +654,25 @@ impl SqsService {
 
         // Move messages to DLQ — also with at-rest bodies untouched,
         // since the DLQ may have its own SSE settings and a downstream
-        // receive call decrypts what's there.
+        // receive call decrypts what's there. If the redrive target ARN
+        // doesn't resolve to a known queue (DLQ deleted, ARN typo), the
+        // message is returned to the source queue instead of being
+        // dropped — real SQS never destroys a message over a
+        // misconfigured redrive target.
         for (dlq_arn, mut msg) in dlq_messages.into_iter() {
+            msg.receipt_handle = None;
+            msg.visible_at = None;
             if let Some(dlq) = state.queues.values_mut().find(|q| q.arn == dlq_arn) {
-                msg.receipt_handle = None;
-                msg.visible_at = None;
                 dlq.messages.push_back(msg);
+            } else {
+                tracing::warn!(
+                    queue = %queue_arn,
+                    dlq_arn = %dlq_arn,
+                    "redrive target queue not found; returning message to source queue"
+                );
+                if let Some(source) = state.queues.values_mut().find(|q| q.arn == queue_arn) {
+                    source.messages.push_back(msg);
+                }
             }
         }
 
--- a/crates/fakecloud-sqs/src/service_tests.rs
+++ b/crates/fakecloud-sqs/src/service_tests.rs
@@ -1232,6 +1232,41 @@ fn list_dead_letter_source_queues_finds_sources() {
     assert!(urls[0].as_str().unwrap().contains("src-q"));
 }
 
+#[test]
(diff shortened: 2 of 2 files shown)


vieiralucas:
Merged — shipping in v0.19.0. Thanks for the clean fix and the red-then-green test. Returning the message to the source queue on an unresolvable DLQ ARN is the right behavior (real SQS keeps redelivering); the minimal runtime fallback over SetQueueAttributes-time validation was a good call.

PR #1681, opened 2026-06-11
fix(dynamodb): evaluate NOT( without a trailing space in condition expressions

### LINKED ISSUE #1680: DynamoDB: NOT(...) ConditionExpression wrongly fails on missing items (no-space NOT)
**Version**: fakecloud 0.18.0 (verified live at `main` @ `ee886ef`; the code path is unchanged on current `main`)

## Summary

A conditional `PutItem` whose `ConditionExpression` wraps `attribute_exists` in `NOT(...)`
raises `ConditionalCheckFailedException` **even when no item exists at the key**. On real
DynamoDB, `attribute_exists` on a missing item is `false`, so `NOT(attribute_exists(pk))`
is `true` and the put succeeds. The plain `attribute_not_exists(pk)` form works correctly
in fakecloud, so the bug is specifically in how `NOT(` is tokenized.

Root cause is in `evaluate_filter_expression`:

```rust
// crates/fakecloud-dynamodb/src/service/helpers/conditions.rs
// Handle NOT prefix (case-insensitive)
if trimmed.len() > 4 && trimmed[..4].eq_ignore_ascii_case("NOT ") {
    return !evaluate_filter_expression(&trimmed[4..], item, expr_attr_names, expr_attr_values);
}
```

It only matches `NOT` with a **trailing space**. An expression of the form `NOT(...)`
(no space) is never recognized as a negation: it isn't split on AND/OR (the operator is
inside the parens), `strip_outer_parens` leaves it untouched (it starts with `N`), it fails
the `"NOT "` check, and it falls through to the single-condition evaluator, which matches
nothing and returns `false`. The result is an unconditional `ConditionalCheckFailedException`
regardless of whether the item exists.

The condition evaluator already models a missing item as an empty map correctly
(`evaluate_condition`), so the missing-item semantics are fine — this is purely the `NOT`
tokenizer. Notably, `extract_function_arg` in the same file already accepts both `func(`
and `func (`; `NOT` should be equally tolerant.

## Reproduction

```sh
EP=http://127.0.0.1:4566
aws dynamodb create-table --endpoint-url $EP --table-name locks \
  --attribute-definitions AttributeName=lock_key,AttributeType=S AttributeName=sort_key,AttributeType=S \
  --key-schema AttributeName=lock_key,KeyType=HASH AttributeName=sort_key,KeyType=RANGE \
  --billing-mode PAY_PER_REQUEST

# A) plain form on a missing key — succeeds (correct)
aws dynamodb put-item --endpoint-url $EP --table-name locks \
  --item '{"lock_key":{"S":"a"},"sort_key":{"S":"-"}}' \
  --condition-expression 'attribute_not_exists(lock_key)'

# B) logically identical NOT form on a missing key — FAILS (incorrect)
aws dynamodb put-item --endpoint-url $EP --table-name locks \
  --item '{"lock_key":{"S":"b"},"sort_key":{"S":"-"}}' \
  --condition-expression 'NOT(attribute_exists(lock_key))'
# => ConditionalCheckFailedException

# C) NOT form on an EXISTING key — fails (correct)
aws dynamodb put-item --endpoint-url $EP --table-name locks \
  --item '{"lock_key":{"S":"a"},"sort_key":{"S":"-"}}' \
  --condition-expression 'NOT(attribute_exists(lock_key) AND attribute_exists(sort_key))'
```

Expected: A passes, B passes, C fails. fakecloud: A passes, **B fails**, C fails.

## Expected

`NOT(...)` (no space) should be tokenized identically to `NOT (...)` / `NOT ...` — recognize
a case-insensitive `NOT` prefix followed by whitespace **or** `(`, then recurse on the
remainder (keeping its parentheses so the inner expression evaluates normally), while not
misparsing an identifier like `NOTanattr`. This mirrors the `func(` / `func (` tolerance
`extract_function_arg` already applies. The committed fix:

```rust
// Handle NOT prefix (case-insensitive). Accept both `NOT (...)` and the
// no-space `NOT(...)` that python_dynamodb_lock and hand-written expressions
// emit — mirroring extract_function_arg's func( / func ( tolerance. Requiring
// the char after `NOT` to be whitespace or `(` avoids misparsing an identifier
// like `NOTanattr`.
if let Some(rest) = trimmed
    .get(..3)
    .filter(|p| p.eq_ignore_ascii_case("NOT"))
    .map(|_| &trimmed[3..])
    .filter(|after| after.starts_with(|c: char| c.is_ascii_whitespace() || c == '('))
{
    return !evaluate_filter_expression(rest, item, expr_attr_names, expr_attr_values);
}
```

## Impact

`python_dynamodb_lock` (AWS's blessed Python distributed-lock library) acquires with
exactly `NOT(attribute_exists(#pk) AND attribute_exists(#sk))`. fakecloud always raises
`ConditionalCheckFailedException`, the library reads that as "lock held", waits out the
lease, and retries forever. In our stack this froze every HTTP handler and queue consumer
that takes an application lock: consumers stopped heartbeating and were liveness-killed in
a loop; handlers hung to client timeout. It presented as a service-level outage while the
emulator looked perfectly healthy.

## Test coverage included with the fix

A regression matrix over condition-expression evaluation against missing items:

- **Unit** — no-space `NOT(attribute_exists(#s))` on missing vs existing items, the
  `NOT(attribute_exists(#pk) AND attribute_exists(#sk))` lock form, and a spaced-form
  regression guard.
- **Corpus** — no-space `NOT(...)` cases added to the SDK-shape expression corpus.
- **E2E** — drives the no-space `NOT(attribute_exists(...))` `PutItem` path end to end
  (succeeds on a missing key, then `ConditionalCheckFailedException` on the now-existing key).

A PR with the fix and these tests is attached.

WarpRat:
## Summary

A conditional `PutItem` whose `ConditionExpression` wraps `attribute_exists` in `NOT(...)`
(no space) raised `ConditionalCheckFailedException` **even when no item exists at the key**.
On real DynamoDB, `attribute_exists` on a missing item is `false`, so `NOT(attribute_exists(pk))`
is `true` and the put should succeed.

The cause is the `NOT`-prefix tokenizer in `evaluate_filter_expression`
(`crates/fakecloud-dynamodb/src/service/helpers/conditions.rs`), which only matched `NOT`
followed by a literal trailing space (`"NOT "`). For `NOT(...)`: it isn't split on AND/OR (the
operator is inside the parens), `strip_outer_parens` leaves it unchanged (it starts with `N`),
it fails the `"NOT "` check, and falls through to the single-condition evaluator, which matches
nothing and returns `false`.

This breaks `python_dynamodb_lock`, which acquires with exactly
`NOT(attribute_exists(#pk) AND attribute_exists(#sk))` — every acquire reads the spurious
failure as "lock held" and retries forever.

Closes #1680.

## Fix

Recognize a case-insensitive `NOT` prefix followed by **whitespace or `(`**, then recurse on
the remainder with its parentheses intact — mirroring the `func(` / `func (` tolerance
`extract_function_arg` already applies in the same file. Requiring the char after `NOT` to be
whitespace or `(` avoids misparsing an identifier like `NOTanattr`; the bare string `"NOT"`
also doesn't match.

## Tests

- **Unit** (`service/tests.rs`, `test_evaluate_condition_not_no_space`): no-space
  `NOT(attribute_exists(#s))` on missing vs existing items, the
  `NOT(attribute_exists(#pk) AND attribute_exists(#sk))` lock form, and a spaced-form
  regression guard.
- **Corpus** (`service/expression_corpus_tests.rs`): no-space `NOT(...)` cases added to the
  SDK-shape expression corpus.
- **E2E** (`fakecloud-e2e/tests/dynamodb.rs`,
  `dynamodb_condition_expression_not_attribute_exists`): `NOT(attribute_exists(id))` succeeds
  on an empty table, then fails with `ConditionalCheckFailedException` on the now-existing key.

## Validation

- `cargo test -p fakecloud-dynamodb` — 285 passed
- `cargo test -p fakecloud-e2e --test dynamodb` — 64 passed
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo fmt --all --check` — clean

---
## Summary by cubic
Fixes parsing of NOT(...) in DynamoDB condition expressions so `NOT(attribute_exists(...))` works without a space, matching AWS behavior and preventing false ConditionalCheckFailedException (affecting `python_dynamodb_lock`).

- Bug Fixes
  - Treat `NOT` (case-insensitive) as a prefix when followed by whitespace or `(`; avoids misparsing identifiers like `NOTanattr`.
  - Recurse on the remainder with parentheses intact; spaced forms remain unchanged.
  - Added unit, corpus, and E2E tests for `NOT(attribute_exists(...))` on missing vs existing items.

<sup>Written for commit 5bf976acdfeec5b8b7bc84a63c719d650b9994de. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/faiscadev/fakecloud/pull/1681?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/fakecloud-dynamodb/src/service/expression_corpus_tests.rs
+++ b/crates/fakecloud-dynamodb/src/service/expression_corpus_tests.rs
@@ -297,6 +297,9 @@ fn filter_sdk_compound_shapes() {
         ),
         ("not-true", "NOT (a = :99)", true),
         ("not-false", "NOT (a = :1)", false),
+        // No-space `NOT(` form (python_dynamodb_lock / hand-written).
+        ("not-true-no-space", "NOT(a = :99)", true),
+        ("not-false-no-space", "NOT(a = :1)", false),
         (
             "and-of-or",
             "((a = :1) OR (a = :99)) AND ((b = :2) OR (b = :99))",
--- a/crates/fakecloud-dynamodb/src/service/helpers/conditions.rs
+++ b/crates/fakecloud-dynamodb/src/service/helpers/conditions.rs
@@ -412,9 +412,18 @@ pub(crate) fn evaluate_filter_expression(
         return evaluate_filter_expression(stripped, item, expr_attr_names, expr_attr_values);
     }
 
-    // Handle NOT prefix (case-insensitive)
-    if trimmed.len() > 4 && trimmed[..4].eq_ignore_ascii_case("NOT ") {
-        return !evaluate_filter_expression(&trimmed[4..], item, expr_attr_names, expr_attr_values);
+    // Handle NOT prefix (case-insensitive). Accept both `NOT (...)` and the
+    // no-space `NOT(...)` that python_dynamodb_lock and hand-written expressions
+    // emit — mirroring extract_function_arg's func( / func ( tolerance. Requiring
+    // the char after `NOT` to be whitespace or `(` avoids misparsing an identifier
+    // like `NOTanattr`.
+    if let Some(rest) = trimmed
+        .get(..3)
+        .filter(|p| p.eq_ignore_ascii_case("NOT"))
+        .map(|_| &trimmed[3..])
+        .filter(|after| after.starts_with(|c: char| c.is_ascii_whitespace() || c == '('))
+    {
+        return !evaluate_filter_expression(rest, item, expr_attr_names, expr_attr_values);
     }
 
     evaluate_single_filter_condition(trimmed, item, expr_attr_names, expr_attr_values)
--- a/crates/fakecloud-dynamodb/src/service/tests.rs
+++ b/crates/fakecloud-dynamodb/src/service/tests.rs
@@ -1819,6 +1819,34 @@ fn test_evaluate_condition_not_operator() {
 }
 
 #[test]
+fn test_evaluate_condition_not_no_space() {
+    // `NOT(` with no space between the keyword and `(` is what
+    // python_dynamodb_lock and most hand-written expressions emit. It must be
+    // tokenized identically to `NOT (...)`.
+    let item = cond_item(&[("state", "active")]);
+    let names = cond_names(&[("#s", "state"), ("#pk", "pk"), ("#sk", "sk")]);
+    let values = cond_values(&[]);
+
+    // NOT(attribute_exists(#s)) on a missing item => NOT false => true
+    assert!(evaluate_condition("NOT(attribute_exists(#s))", None, &names, &values).is_ok());
+    // ...on an existing item => NOT true => false
+    assert!(evaluate_condition("NOT(attribute_exists(#s))", Some(&item), &names, &values).is_err());
+
+    // The python_dynamodb_lock acquire form on a missing item => passes.
+    assert!(evaluate_condition(
+        "NOT(attribute_exists(#pk) AND attribute_exists(#sk))",
+        None,
+        &names,
+        &values
+    )
+    .is_ok());
+
+    // Regression guard: the spaced form still behaves as before.
+    assert!(evaluate_condition("NOT (attribute_exists(#s))", None, &names, &values).is_ok());
+    assert!(evaluate_condition("NOT attribute_exists(#s)", Some(&item), &names, &values).is_err());
+}
+
+#[test]
 fn test_evaluate_condition_begins_with() {
     // After unification, conditions support begins_with via
     // evaluate_single_filter_condition (previously only filters had it).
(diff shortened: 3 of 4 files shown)


vieiralucas:
Merged — shipping in v0.19.0. The no-space `NOT(` repro via python_dynamodb_lock was spot on, and mirroring `extract_function_arg`'s `func(`/`func (` tolerance kept it consistent. Thanks!

PR #1687, opened 2026-06-11
fix(dynamodb): support legacy KeyConditions/QueryFilter Query parameters

### LINKED ISSUE #1685: DynamoDB Query/Scan reject legacy KeyConditions/QueryFilter/ScanFilter parameters
**Version**: fakecloud 0.18.0 (reproduced against current `main`)

## Summary

DynamoDB `Query` only implements the expression API (`KeyConditionExpression`). The legacy
`KeyConditions` / `QueryFilter` parameters are ignored, and the request fails with an error
that ironically names the supplied parameter as missing:

```
ValidationException: Either the KeyConditions or KeyConditionExpression
parameter must be specified in the request.
```

`Query` (`crates/fakecloud-dynamodb/src/service/queries.rs`) reads only
`KeyConditionExpression` and rejects the request when it's absent — it never inspects
`KeyConditions`/`QueryFilter`.

**On the obvious objection — "these were deprecated in 2014":** yes, AWS deprecated
`KeyConditions`/`QueryFilter` in 2014 in favor of the expression API. But AWS never removed
them: real DynamoDB and every official SDK still accept them today, and they're common in
long-lived codebases and pinned-SDK call sites (AWS never forced a migration). Since
fakecloud's whole premise is high-fidelity AWS behavior, silently rejecting input that real
DynamoDB accepts is a parity bug regardless of the parameter's age.

## Reproduction

```sh
EP=http://127.0.0.1:4566
aws --region us-east-1 dynamodb create-table --endpoint-url $EP --table-name t \
  --attribute-definitions AttributeName=pk,AttributeType=S \
  --key-schema AttributeName=pk,KeyType=HASH --billing-mode PAY_PER_REQUEST
aws --region us-east-1 dynamodb put-item --endpoint-url $EP --table-name t \
  --item '{"pk":{"S":"a"},"v":{"S":"1"}}'
aws --region us-east-1 dynamodb query --endpoint-url $EP --table-name t \
  --key-conditions '{"pk":{"AttributeValueList":[{"S":"a"}],"ComparisonOperator":"EQ"}}'
# actual:   ValidationException: Either the KeyConditions or KeyConditionExpression ...
# expected: returns the item {"pk":"a","v":"1"}
```

## Expected

`Query` accepts the legacy `KeyConditions` (and `QueryFilter`) parameters and returns the
same result as the equivalent `KeyConditionExpression`. The clean fix is to **translate at
the request boundary** — each `{AttributeValueList, ComparisonOperator}` maps mechanically
onto its expression equivalent, then reuse the existing expression evaluator:

- `EQ` → `=`, `LE` → `<=`, `LT` → `<`, `GE` → `>=`, `GT` → `>`
- `BEGINS_WITH` → `begins_with(...)`
- `BETWEEN` → `BETWEEN`, entries AND-ed together

Ideally also `QueryFilter` → `FilterExpression`, and the rest of the legacy family
(`ScanFilter`, `Expected`, `AttributesToGet`). If full support is out of scope, at minimum
return an honest "legacy KeyConditions is not supported" error rather than the current one,
which names the very parameter the caller supplied.

## Impact

Any pre-2014-style DynamoDB code — common in long-lived codebases because AWS never forced
migration — silently can't read through fakecloud. This broke two of our services' read
paths and surfaced as e2e timeouts in *different* services; it took a long session to trace
back to the rejected legacy parameter.

I've opened a PR implementing this: it translates the legacy parameters at the request
boundary and reuses the existing expression evaluator, covering `KeyConditions` +
`QueryFilter` (Query) and `ScanFilter` (Scan), with `Expected` / `AttributesToGet` called out
as follow-ups. See the linked PR below.

WarpRat:
Closes #1685.

## What

`Query` and `Scan` only implemented the expression API (`KeyConditionExpression` /
`FilterExpression`). The legacy `KeyConditions` / `QueryFilter` / `ScanFilter` parameters were
ignored, and `Query` rejected the request with a message that ironically named the parameter
the caller *did* supply:

```
ValidationException: Either the KeyConditions or KeyConditionExpression
parameter must be specified in the request.
```

## Why (parity, not endorsement)

AWS deprecated `KeyConditions`/`QueryFilter` in **2014** in favor of the expression API, but
never removed them — real DynamoDB and every official SDK still accept them today, and they
remain common in long-lived and pinned-SDK codebases (AWS never forced a migration). Since
fakecloud's premise is high-fidelity AWS behavior, silently rejecting input real DynamoDB
accepts is a parity gap regardless of the parameter's age.

## How

Translate the legacy form at the request boundary into the equivalent expression string —
hoisting attribute names into `#`-placeholders and values into `:`-placeholders in the
existing expr-attr maps — then feed it through the existing `evaluate_key_condition` /
`evaluate_filter_expression` machinery. **No evaluation logic changes.**

Operator mapping: `EQ`→`=`, `LE`→`<=`, `LT`→`<`, `GE`→`>=`, `GT`→`>`, `NE`→`<>`,
`BEGINS_WITH`→`begins_with(...)`, `BETWEEN`→`x BETWEEN :lo AND :hi`, `CONTAINS`/`NOT_CONTAINS`,
`NULL`/`NOT_NULL`→`attribute_not_exists`/`attribute_exists`, `IN`. Entries are AND-ed. Mixing
the legacy and expression forms for the same role is rejected (matching AWS); an absent key
condition keeps the original error.

## Scope

Covers `KeyConditions` + `QueryFilter` (Query) and `ScanFilter` (Scan). `Expected` (conditional
writes) and `AttributesToGet` (legacy projection) are the same family and a sensible follow-up,
but deferred here to keep the change focused.

## Tests

- **Unit** (`conditions.rs`): the translation helper directly — `EQ` + `BEGINS_WITH`, `BETWEEN`,
  filter-only operators, and error cases (unknown operator, wrong arg count, key-role rejecting
  a filter-only operator).
- **E2E** (`dynamodb.rs`, via the `aws` CLI, which emits the legacy wire shape): the canonical
  `--key-conditions` repro, a composite `EQ`+`BEGINS_WITH` query, `--query-filter`,
  `--scan-filter`, and the mixed-form rejection.

Validated: `cargo test -p fakecloud-dynamodb` (290 ok), `cargo test -p fakecloud-e2e dynamodb`
(68 ok), `cargo clippy --workspace --all-targets -- -D warnings` (clean), `cargo fmt --all --check` (clean).

---
## Summary by cubic
Add support for DynamoDB legacy `KeyConditions`, `QueryFilter`, and `ScanFilter` by translating them into expression syntax. Restores AWS parity for Query/Scan and fixes rejections of valid legacy requests; evaluation logic is unchanged.

- **Bug Fixes**
  - Query/Scan now accept legacy params and translate them to `KeyConditionExpression`/`FilterExpression` with placeholders, then evaluate via existing code.
  - Mixing legacy and expression forms for the same role is rejected with a `ValidationException`, matching AWS.
  - Query still requires a key condition; missing conditions are rejected (no silent full-table scan).
  - Added unit and E2E tests (via `aws` CLI) covering key conditions, non-key filters, and mixed-form rejection.

<sup>Written for commit ed1aa290b08101c8b8c5118cb8727f4317a8159a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/faiscadev/fakecloud/pull/1687?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/fakecloud-dynamodb/src/service/helpers/conditions.rs
+++ b/crates/fakecloud-dynamodb/src/service/helpers/conditions.rs
@@ -382,6 +382,178 @@ pub(crate) fn evaluate_size_comparison(
     })
 }
 
+/// Which legacy condition family is being translated. `KeyConditions`
+/// (the `Query` key) accept only the comparison + range operators a key
+/// condition allows; the filter families (`QueryFilter` / `ScanFilter`)
+/// accept the broader operator set the FilterExpression evaluator supports.
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub(crate) enum LegacyConditionRole {
+    Key,
+    Filter,
+}
+
+/// Translate a legacy `KeyConditions` / `QueryFilter` / `ScanFilter` map into
+/// the equivalent expression string, hoisting attribute names into `#leg*`
+/// placeholders and values into `:legv*` placeholders in the supplied
+/// expr-attr maps so the result can be fed straight through the existing
+/// `evaluate_key_condition` / `evaluate_filter_expression` machinery.
+///
+/// AWS deprecated these parameters in 2014 in favor of the expression API,
+/// but real DynamoDB and every SDK still accept them — this is a parity
+/// translation, not new evaluation logic. Entries are AND-ed together
+/// (the only join AWS supports for the legacy form). Iteration is in sorted
+/// attribute-name order so the synthesized expression is deterministic.
+pub(crate) fn translate_legacy_conditions(
+    conditions: &serde_json::Map<String, Value>,
+    role: LegacyConditionRole,
+    names: &mut HashMap<String, String>,
+    values: &mut HashMap<String, Value>,
+) -> Result<String, AwsServiceError> {
+    // Role-specific placeholder tag so the key-condition and filter
+    // translations in a single Query don't clobber each other's entries in
+    // the shared expr-attr maps.
+    let tag = match role {
+        LegacyConditionRole::Key => "k",
+        LegacyConditionRole::Filter => "f",
+    };
+
+    let mut entries: Vec<(&String, &Value)> = conditions.iter().collect();
+    entries.sort_by(|a, b| a.0.cmp(b.0));
+
+    let mut fragments: Vec<String> = Vec::with_capacity(entries.len());
+    for (idx, (attr, spec)) in entries.iter().enumerate() {
+        let operator = spec
+            .get("ComparisonOperator")
+            .and_then(|v| v.as_str())
+            .ok_or_else(|| {
+                legacy_validation_err(format!(
+                    "ComparisonOperator is required for the condition on attribute {attr}"
+                ))
+            })?;
+        let value_list: &[Value] = spec
+            .get("AttributeValueList")
+            .and_then(|v| v.as_array())
+            .map(|a| a.as_slice())
+            .unwrap_or(&[]);
+
+        // Hoist the attribute name into a collision-safe placeholder.
+        let name_ph = format!("#leg{tag}{idx}");
+        names.insert(name_ph.clone(), (*attr).clone());
+
+        // Hoist each operand value into its own placeholder.
+        let mut value_phs: Vec<String> = Vec::with_capacity(value_list.len());
+        for (vi, v) in value_list.iter().enumerate() {
+            let ph = format!(":legv{tag}{idx}_{vi}");
+            values.insert(ph.clone(), v.clone());
+            value_phs.push(ph);
+        }
+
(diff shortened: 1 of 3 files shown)


vieiralucas:
Merged — shipping in v0.19.0. Correct parity call: AWS never removed the 2014 legacy `KeyConditions`/`QueryFilter`/`ScanFilter`, so accepting them matches real DynamoDB. Translating at the request boundary into the existing evaluator was clean. Thanks.

PR #1711, opened 2026-06-14
fix(stepfunctions): faithful lambda:invoke optimized integration

WarpRat:
**Version**: fakecloud (current `main` @ `d828a709`)

## Summary

The Step Functions `arn:aws:states:::lambda:invoke` optimized integration diverges
from AWS in three related ways, all in `invoke_lambda_direct` and the
`lambda:invoke` dispatch arm in
`crates/fakecloud-stepfunctions/src/interpreter.rs`:

1. **Lambda function errors are reported as task success.** The Lambda runtime
   returns HTTP 200 with an `{errorMessage, errorType, stackTrace}` body for an
   unhandled function error. The interpreter treats that as a successful task
   result, so the function's `errorType` never becomes the task's Error — `Retry`
   and `Catch` clauses keyed on it (e.g. a custom `RetryableServerException`) never
   fire, and the workflow proceeds on a silently-failed step.

2. **Invocation/transport failures map to `States.TaskFailed`.** When the invoke
   itself fails (transport error, function not found, etc.), the task fails with
   `States.TaskFailed`. AWS surfaces these as `Lambda.*` error names
   (`Lambda.ServiceException`, `Lambda.Unknown`, …), which are what `Retry` rules
   are written against — so retryable failures aren't retried.

3. **Successful results are not wrapped in the AWS response envelope.** The
   optimized integration returns the bare function payload instead of
   `{ExecutedVersion, Payload, StatusCode}`. Templates that select `$.Payload`
   (the AWS-documented shape) get `null`, corrupting every downstream step that
   reads the previous task's output.

## Reproduction

```sh
EP=http://127.0.0.1:4566
# A Lambda whose handler raises a custom exception (returns 200 + errorType body).
# A state machine Task invoking it via the optimized integration with a Retry on
# the custom error name and ResultSelector: { "v.$": "$.Payload" }.
aws --endpoint-url $EP stepfunctions get-execution-history --execution-arn <arn>
```

Observed vs AWS:
- **Function error:** fakecloud → `TaskSucceeded` with the error JSON as output;
  AWS → `TaskFailed` with `error = <errorType>`, `Retry`/`Catch` fire.
- **Transport failure:** fakecloud → `States.TaskFailed`; AWS → `Lambda.Unknown`
  (retryable).
- **Success envelope:** fakecloud → output is the bare payload, `$.Payload` →
  `null`; AWS → output is `{ExecutedVersion, Payload, StatusCode}`.

## Expected

The optimized `lambda:invoke` integration matches AWS:
- A 200 response carrying `errorType`/`errorMessage` fails the task with the
  function's `errorType` as the Error and the payload as the Cause (so `Retry`/
  `Catch` match it).
- Invoke-layer failures surface as `Lambda.*` error names.
- Successful task output is `{ExecutedVersion, Payload, StatusCode}`; error results
  are not wrapped (matching AWS, where a failed `lambda:invoke` task has no result
  envelope).

## Impact

Any SAM/Step Functions app that fans work out through a dispatcher Lambda is
affected: function errors are swallowed, retryable failures aren't retried, and
`$.Payload`-based result wiring breaks — so multi-step workflows silently produce
wrong/empty output instead of failing or retrying as designed.

## The fix

Three focused commits on the `lambda:invoke` path in `interpreter.rs`:

1. **Function errors → task errors.** In `invoke_lambda_direct`, a 200 payload
   carrying **both** `errorType` and `errorMessage` now fails the task with that
   `errorType` as the Error and the serialized payload as the Cause. Requiring both
   keys avoids misreading a legitimate success payload that merely contains one of
   those field names.
2. **Transport failures → `Lambda.*`.** Invoke-layer failures map to
   `Lambda.Unknown` (the honest default for an unstructured delivery error), with
   `Function not found` surfaced as `Lambda.ResourceNotFoundException`.
3. **Success envelope.** The optimized-integration dispatch arm wraps a successful
   result in `{ExecutedVersion: "$LATEST", Payload, StatusCode: 200}`. Error
   results pass through unwrapped (AWS has no result envelope on a failed task),
   and the direct-ARN path intentionally stays unwrapped (bare payload, matching
   AWS).

They edit the same arm and share one test stub, so they ship together.

## Test plan

- `cargo test -p fakecloud-stepfunctions` — 173 passed. New unit cases cover:
  function-error→task-error, Retry-on-custom-errorType fires, transport→
  `Lambda.Unknown`, `$.Payload` envelope, non-error passthrough.
- `cargo test -p fakecloud-e2e --test stepfunctions --test cloudformation_stepfunctions --test stepfunctions_task_token --test stepfunctions_persistence`
  — 77 passed, 0 failed, 2 ignored.

---
## Summary by cubic
Aligns the Step Functions `arn:aws:states:::lambda:invoke` optimized integration with AWS so retries and result paths work correctly. Function errors now fail the task, invoke-layer failures use `Lambda.*` error names, and successful results return the AWS response envelope.

- **Bug Fixes**
  - Function error payloads (200 with `errorType` and `errorMessage`) now fail the task: Error=`<errorType>`, Cause=payload; `Retry`/`Catch` on custom errors now work.
  - Invoke/transport failures map to AWS names: default `Lambda.Unknown`; `Function not found` → `Lambda.ResourceNotFoundException`.
  - Successful results are wrapped as `{ExecutedVersion: "$LATEST", Payload, StatusCode: 200}`; direct-ARN invocations stay unwrapped; error results are not wrapped.

<sup>Written for commit ab4e36d3b066925b65de074798b1c8321d13f136. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/faiscadev/fakecloud/pull/1711?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/fakecloud-stepfunctions/src/interpreter.rs
+++ b/crates/fakecloud-stepfunctions/src/interpreter.rs
@@ -1862,6 +1862,33 @@ async fn invoke_lambda_direct(
             let response_str = String::from_utf8_lossy(&bytes);
             let value: Value =
                 serde_json::from_str(&response_str).unwrap_or(json!(response_str.to_string()));
+
+            // The Lambda runtime returns HTTP 200 even for unhandled function
+            // errors, encoding the failure in the payload as
+            // `{"errorMessage", "errorType", "stackTrace"}`. The delivery trait
+            // only conveys bytes, so the payload is the only available signal —
+            // the same heuristic the Lambda service uses for async destination
+            // routing. Mirror real AWS: a function error fails the task with the
+            // function's `errorType` as the Error and the serialized payload as
+            // the Cause, which is exactly what makes Retry/Catch on custom
+            // exception names work.
+            //
+            // Require BOTH `errorType` and `errorMessage` (not either) to avoid
+            // misreading a legitimate success payload that merely happens to
+            // contain one of those field names as a function error.
+            if let Some(obj) = value.as_object() {
+                if obj.contains_key("errorType") && obj.contains_key("errorMessage") {
+                    let error_type = obj
+                        .get("errorType")
+                        .and_then(Value::as_str)
+                        .unwrap_or("Exception")
+                        .to_string();
+                    let cause = serde_json::to_string(&value)
+                        .expect("serde_json::Value serialization is infallible");
+                    return Err((error_type, cause));
+                }
+            }
+
             Ok(value)
         }
         Some(Err(e)) => Err(("States.TaskFailed".to_string(), e)),
--- a/crates/fakecloud-stepfunctions/src/interpreter.rs
+++ b/crates/fakecloud-stepfunctions/src/interpreter.rs
@@ -1096,7 +1096,24 @@ async fn invoke_resource(
         } else {
             input.clone()
         };
-        return invoke_lambda_direct(function_name, &payload, delivery, timeout_seconds).await;
+        // The optimized `lambda:invoke` integration returns the AWS Invoke API
+        // response envelope, not the bare function output — `ResultSelector`
+        // expressions like `{"value.$": "$.Payload"}` depend on it. Wrap only
+        // the successful result; error results (function errors and transport
+        // failures) pass through unchanged, since a failed lambda:invoke task
+        // has no result envelope on AWS. The direct-ARN path above
+        // intentionally stays unwrapped — there real AWS returns the bare
+        // payload. `SdkHttpMetadata`/`SdkResponseMetadata` are omitted; real
+        // templates select `Payload`/`StatusCode`.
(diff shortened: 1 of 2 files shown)


WarpRat:
Tracking issue: #1712

vieiralucas:
Thanks @WarpRat! Reviewed the `lambda:invoke` optimized-integration semantics against real Step Functions — function-error -> task failure, `Lambda.*` error names, and the `Payload`/`StatusCode` envelope all check out, with solid unit coverage. Merged as-is.

PR #1713, opened 2026-06-14
fix(lambda): warm-pool unique pod names + fast dead-instance failover

WarpRat:
**Version**: fakecloud (current `main` @ `d828a709`), Kubernetes Lambda backend
(`FAKECLOUD_LAMBDA_BACKEND=k8s`).

## Summary

Under concurrent invocations of a single function, the k8s Lambda warm pool
thrashes a **single deterministically-named pod** and invocations fail with
`Lambda.Unknown`. The warm pool runs one in-flight invocation per instance (the
RIE can't overlap), but the pod is named deterministically per function+deploy —
so when concurrency requires a second instance, or a cold start recreates an
instance whose previous pod is still terminating, the k8s API rejects it:

```
container failed to start: k8s create pod: object is being deleted:
  pods "fakecloud-lambda-<fn>-<deployid>" already exists: AlreadyExists
... or ...
  pods "fakecloud-lambda-<fn>-<deployid>" not found: NotFound
```

The failed launch surfaces as `Lambda.Unknown` to the caller. A separate facet:
when a warm pod's IP is unreachable (terminating), the invoke can hang on the full
function timeout before failing, rather than failing over fast.

## Reproduction

With the k8s backend, drive **concurrent** `Invoke` calls at one function (e.g. a
Step Functions workflow that fans out several `lambda:invoke` tasks at once).
`kubectl get events` for the function's pod shows the same pod name being
Killed and recreated repeatedly within seconds (no node eviction involved), and
some invocations fail with `Lambda.Unknown: ... object is being deleted ...
AlreadyExists` / `... not found`.

## Expected

Concurrent invocations are served without create/delete races:
- Each launched instance gets a **unique pod name** (so a terminating old pod
  never blocks a replacement, and the pool can hold more than one live pod per
  function under load).
- A dead/unreachable warm instance is detected and failed over **fast** (a short
  reachability check), not after the full function timeout.

## Impact

Any workload that invokes one function concurrently (Step Functions fan-out, an
event burst, parallel test suites) sees a high rate of spurious `Lambda.Unknown`
failures under the k8s backend. In our bake-off this made every multi-step
workflow fail reliably under load, presenting as widespread "flaky" test failures.

## The fix

- **Unique pod names.** `runtime/k8s/spec.rs` gains `unique_pod_name`, which salts
  the hashed function+deploy id with a process-monotonic atomic counter to produce
  a fresh, length-bounded, DNS-1123-safe name per launch; `runtime/k8s/mod.rs`
  overrides the pod's metadata name with it. The pool already stored distinct
  `WarmEntry` instances, so no pool-keying change was needed — it can now actually
  hold >1 live pod per function.
- **Fast failover.** `runtime/facade.rs` adds `endpoint_reachable` — a ~1.5s TCP
  reachability probe run before forwarding the payload (and a matching reqwest
  `connect_timeout` backstop). A black-holed pod IP is detected in ~1s and failed
  over to a cold start, instead of hanging `~func.timeout + 5s` (~305s). The
  streaming path gets the same probe. The connect-only retry budget is raised to
  five cold-start replacements (`MAX_INVOKE_ATTEMPTS`).

## Design notes / alternatives considered

- **Unique pod names vs. wait-for-delete / retry-on-`AlreadyExists`.** Both
  alternatives fix the recreate race, but neither lifts the **concurrency ceiling**:
  a deterministic name means at most one live pod per function, which is the actual
  failure under fan-out load. Unique names fix the race *and* let the pool run
  multiple concurrent instances — so that's the chosen approach. Names are salted
  with a process-monotonic counter (not a clock or RNG) to stay deterministic and
  collision-free without a wall-clock dependency.
- **Why the retry is safe.** Failover only happens on a failed TCP connect (probe
  or `e.is_connect()`). A refused/black-holed connection provably never reached the
  handler, so retrying cannot double-execute side effects. Reset/timeout failures
  mid-flight still surface immediately, unchanged.
- **Probe timeout (1.5s).** Long enough to tolerate a slow-but-live pod's TCP
  handshake, short enough that a dead instance fails over well inside a state
  machine's retry window rather than consuming the full invoke timeout. Tunable via
  the `REACHABILITY_PROBE_TIMEOUT` constant.
- **Pool keying/limits unchanged.** No change to how the pool keys or caps
  instances; `DEFAULT_MAX_CONCURRENCY` (10) still governs reuse-vs-cold-start.

## Test plan

- `cargo test -p fakecloud-lambda` — warm-pool tests extended:
  `dead_instance_is_evicted_and_retried` now exercises two dead instances in a row,
  and a new `black_holed_instance_fails_over_fast` asserts a black-holed endpoint
  fails over fast (no multi-minute hang). The mock RIE now counts only connections
  that send a request, so a bare probe isn't miscounted as an invocation. Timing
  assertions avoid wall-clock per the repo constraint.

---
## Summary by cubic
Fixes flaky k8s Lambda warm-pool invocations under concurrent load by giving each instance a unique Pod name and adding a fast reachability probe to evict dead pods. This removes create/delete races and cuts dead-instance hangs from ~function timeout to ~1s.

- **Bug Fixes**
  - Unique pod names: new `unique_pod_name(...)` salts function+deploy with a monotonic counter and overrides `metadata.name`. Prevents `AlreadyExists` races and allows multiple live Pods per function.
  - Fast failover: pre-invoke TCP probe (~1.5s) plus `reqwest` `connect_timeout`. Unreachable warm Pods are evicted and replaced; retry budget `MAX_INVOKE_ATTEMPTS = 5`. Streaming path included.
  - Safe retries: only on failed connects (never reached handler). Mid-flight errors still return immediately.
  - Tests: add black-holed endpoint test; extend dead-instance retries; mock RIE ignores bare connect probes; add `unique_pod_name` tests for uniqueness and DNS-1123 safety.

<sup>Written for commit 50a01ba5eaf599a6b86066aefef3ce79b136b35f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/faiscadev/fakecloud/pull/1713?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/fakecloud-lambda/src/runtime/facade.rs
+++ b/crates/fakecloud-lambda/src/runtime/facade.rs
@@ -44,6 +44,18 @@ pub(crate) struct WarmEntry {
 /// on a busy instance rather than starting a new one.
 const DEFAULT_MAX_CONCURRENCY: usize = 10;
 
+/// Max attempts per invocation when a reserved warm instance turns out to be
+/// unreachable. Each failover is a fast probe (or connect error) plus a cold
+/// start; the connection provably never reached the handler, so retrying can't
+/// double-execute. Bounded so an all-dead pool can't spin forever.
+const MAX_INVOKE_ATTEMPTS: u32 = 5;
+
+/// Timeout for the pre-invoke TCP reachability probe. A black-holed Pod IP (a
+/// killed pod / drained node that drops packets with no RST) would otherwise
+/// hang the full invoke timeout (~`func.timeout + 5s`); this detects it in ~1s
+/// so the state-machine retry can succeed within its window.
+const REACHABILITY_PROBE_TIMEOUT: Duration = Duration::from_millis(1500);
+
 /// A reserved invocation slot: a warm instance plus the held busy guard
 /// that grants exclusive use of it until the guard drops.
 struct Slot {
@@ -77,6 +89,18 @@ fn deploy_id_from(code_sha256: &str, layers: &[Vec<u8>]) -> String {
         hasher.update(layer_hasher.finalize());
     }
     base64::engine::general_purpose::URL_SAFE_NO_PAD.encode(hasher.finalize())
+}
+
+/// Quick liveness check: can a TCP connection to `endpoint` (`host:port`) be
+/// opened within `timeout`? Used before forwarding a payload to a warm instance
+/// so a dead/black-holed Pod is detected in ~1s instead of hanging the full
+/// invoke timeout. A failed connect provably never reached the handler, so the
+/// caller can safely evict and retry.
+async fn endpoint_reachable(endpoint: &str, timeout: Duration) -> bool {
+    matches!(
+        tokio::time::timeout(timeout, tokio::net::TcpStream::connect(endpoint)).await,
+        Ok(Ok(_))
+    )
 }
 
 pub struct LambdaRuntime {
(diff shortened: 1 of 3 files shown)


WarpRat:
Tracking issue: #1714

vieiralucas:
Thanks @WarpRat! The unique-pod-name + fast-failover fix is solid. I pushed three small follow-throughs to your branch before merging: a `unique_pod_name` unit test (uniqueness + DNS-1123 safety + <=63 chars), an ASCII `->` in one comment, and corrected the "up to twice" doc to match `MAX_INVOKE_ATTEMPTS = 5`. Merged.

PR #1717, opened 2026-06-14
fix(cloudformation): stack events for changeset-created stacks

WarpRat:
**Version**: fakecloud (current `main`/`upstream/main` @ `d828a709`)

## Summary

Two related stack-event defects break the SAM CLI's deploy-progress wait:

1. **`CreateChangeSet` records no event when it mints a new stack.** A `CREATE`
   changeset creates the stack in `REVIEW_IN_PROGRESS` but logs no stack event, so
   `DescribeStackEvents` is empty until `ExecuteChangeSet`. `sam deploy` calls
   `get_last_event_time` **before** executing — it does
   `describe_stack_events(...)["StackEvents"][0]["Timestamp"]` — and crashes:
   ```
   Error: list index out of range   (samcli/lib/deploy/deployer.py get_last_event_time)
   ```
   AWS emits a `REVIEW_IN_PROGRESS` stack event when the review stack is created,
   so the list is never empty.

2. **Stack event timestamps are whole-second.** Events recorded within the same
   second share an identical timestamp. `sam`'s wait marks the last-seen event and
   advances only on events with a **strictly later** timestamp; a stack that
   provisions within one second has its terminal `CREATE_COMPLETE` at the same
   timestamp as the marker, so sam never registers completion and hangs until the
   job times out.

These compose: fix (1) so the list isn't empty, and (2) so the recorded events are
distinguishable — together they make `sam deploy`'s wait work on fast stacks.

## Reproduction

```sh
EP=http://127.0.0.1:4566
cat > /tmp/cs.yaml <<'EOF'
Resources: { Q: { Type: "AWS::SQS::Queue" } }
EOF
aws --region us-east-1 cloudformation create-change-set --endpoint-url $EP \
  --stack-name cs-ev --change-set-name cs1 --change-set-type CREATE \
  --template-body file:///tmp/cs.yaml
# BEFORE: empty. AFTER: one REVIEW_IN_PROGRESS event.
aws --region us-east-1 cloudformation describe-stack-events --endpoint-url $EP \
  --stack-name cs-ev
aws --region us-east-1 cloudformation execute-change-set --endpoint-url $EP \
  --stack-name cs-ev --change-set-name cs1
# AFTER: event Timestamps are strictly increasing (sub-second), REVIEW_IN_PROGRESS
# < ... < CREATE_COMPLETE, even though it all happened within one second.
aws --region us-east-1 cloudformation describe-stack-events --endpoint-url $EP \
  --stack-name cs-ev --query 'StackEvents[].Timestamp'
```

## Expected

- Creating a stack via `CreateChangeSet` emits a `REVIEW_IN_PROGRESS`
  `AWS::CloudFormation::Stack` event immediately.
- Stack event timestamps are sub-second and strictly monotonic within a stack, so
  consecutive events (esp. the pre-execute marker vs. the terminal event) are
  distinguishable.

## Impact

`sam deploy` against fakecloud either crashes with `IndexError` (empty events
before execute) or hangs forever waiting for completion on any stack that
provisions within one second. Both make plain `sam deploy` unusable.

Happy to test a branch or send a PR.

## The fix

Two commits, shipped together because the second is what makes the first usable on
fast stacks:

1. **Record `REVIEW_IN_PROGRESS` on changeset-created stacks.** The `CreateChangeSet`
   new-stack path in `crates/fakecloud-cloudformation/src/extras.rs` now records an
   initial `REVIEW_IN_PROGRESS` `AWS::CloudFormation::Stack` event via the existing
   `record_stack_status_event` helper, so `DescribeStackEvents` is non-empty before
   `ExecuteChangeSet` (sam's `get_last_event_time` indexes `[0]`).
2. **Sub-second, strictly-monotonic timestamps.** `record_stack_*_event` in
   `service.rs` now stamps millisecond-precision timestamps and bumps by 1ms when
   the clock hasn't advanced past the previous event, and serializes fractional
   seconds. The precision trap is handled: `now` is truncated to the stored
   (millisecond) resolution **before** comparing against the previous value, so the
   strict-ordering guarantee matches what's actually serialized (two events in the
   same millisecond would otherwise serialize identically despite `now > prev` at
   full precision). This makes a fast changeset's `REVIEW_IN_PROGRESS` marker
   strictly earlier than its terminal `CREATE_COMPLETE`.

## Test plan

- `cargo test -p fakecloud-cloudformation` — 215 passed, including
  `create_change_set_records_review_in_progress_event` (≥1 `REVIEW_IN_PROGRESS`
  event) and `changeset_stack_events_have_monotonic_subsecond_timestamps` (full
  create lifecycle within one second, strictly-increasing sub-second timestamps).
  Timing assertions avoid wall-clock per the repo constraint.
- `cargo test -p fakecloud-e2e --test cloudformation_execute_change_set --test cloudformation` — 6 + 15 passed.

---
## Summary by cubic
Fixes CloudFormation stack events for stacks created via CreateChangeSet by emitting a `REVIEW_IN_PROGRESS` event and using millisecond, strictly increasing timestamps. This prevents `sam deploy` from crashing on empty events and from hanging on fast stacks.

- **Bug Fixes**
  - Record a `REVIEW_IN_PROGRESS` `AWS::CloudFormation::Stack` event when a CREATE-type change set creates a new stack, so `DescribeStackEvents` is not empty before `ExecuteChangeSet`.
  - Stamp stack events with millisecond precision and guarantee strictly monotonic ordering (bump by 1ms when needed), with fractional seconds serialized in responses.

<sup>Written for commit 106499dbdcd6eae38dbacc7f17395e26c976861d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/faiscadev/fakecloud/pull/1717?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/fakecloud-cloudformation/src/extras.rs
+++ b/crates/fakecloud-cloudformation/src/extras.rs
@@ -534,6 +534,21 @@ impl CloudFormationService {
                                     outputs: Vec::new(),
                                 },
                             );
+                            // Real CloudFormation emits a stack event when a
+                            // CREATE change set puts a new stack into
+                            // `REVIEW_IN_PROGRESS`. Record it so the event log
+                            // is non-empty between CreateChangeSet and
+                            // ExecuteChangeSet: `sam deploy` reads the most
+                            // recent event as a marker (`get_last_event_time`)
+                            // and unconditionally indexes `[0]`, throwing an
+                            // `IndexError` if the list is empty.
+                            crate::service::record_stack_status_event(
+                                state,
+                                &stack_id_str,
+                                &stack_name,
+                                "AWS::CloudFormation::Stack",
+                                "REVIEW_IN_PROGRESS",
+                            );
                         }
                         // Already in review (an earlier CREATE change set):
                         // additional CREATE change sets are allowed; keep the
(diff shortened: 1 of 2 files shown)


WarpRat:
Tracking issue: #1718

vieiralucas:
Thanks @WarpRat! Both fixes verified — the `REVIEW_IN_PROGRESS` event on CreateChangeSet and the millisecond, strictly-monotonic timestamps fix the exact `sam deploy` hang/crash. Clean and well-tested. Merged.

PR #1719, opened 2026-06-14
fix(cloudformation): dependency-ordered resource provisioning

WarpRat:
**Version**: fakecloud (current `main`/`upstream/main` @ `d828a709`)

## Summary

CloudFormation provisions a stack's resources in iteration order, with no
topological ordering over `DependsOn` / `Ref` / `Fn::GetAtt` / `Fn::Sub`. When a
resource is provisioned **before** another resource it references, the reference
resolves to the **logical ID** (a placeholder) instead of the target's physical
name — and that placeholder is baked into the dependent resource permanently.

Concretely: a state machine whose `DefinitionSubstitutions` reference a Lambda via
`{"Ref": "<FnLogicalId>"}` ends up with the **logical ID** substituted into its
ASL when the state machine happens to provision first. At runtime every task fails:

```
States.TaskFailed → Lambda.ResourceNotFoundException: Function not found: <FnLogicalId>
```

`resolve_refs_full` (`crates/fakecloud-cloudformation/src/template/intrinsics.rs`,
~lines 190–199) returns the physical ID only if the target is already provisioned;
otherwise it falls back to the logical ID "and lets incremental provisioning
rewrite it" — but for values baked into a provisioned resource (a state machine
definition), nothing rewrites them, and there's no provisioning order that would
have made the target exist first.

## Reproduction

```sh
EP=http://127.0.0.1:4566
# Template where a StateMachine (declared FIRST) references a Function via
# DefinitionSubstitutions { fn: !Ref MyFn }, FunctionName fixed to e.g. my-fn-1.
aws --region us-east-1 cloudformation create-stack --endpoint-url $EP \
  --stack-name dep-order --template-body file:///tmp/dep.yaml --capabilities CAPABILITY_IAM
# Inspect the deployed definition:
aws --region us-east-1 stepfunctions describe-state-machine --endpoint-url $EP \
  --state-machine-arn <arn> --query 'definition'
# actual:   FunctionName resolves to "MyFn" (logical id)
# expected: FunctionName resolves to "my-fn-1" (the function's name)
```

## Expected

Resources provision in dependency order (topological sort over `DependsOn` plus
implicit `Ref`/`Fn::GetAtt`/`Fn::Sub` edges), so a referenced resource exists
before its referrer resolves — and references resolve to physical names/ARNs, not
logical IDs. (Real CloudFormation does this; it also rejects dependency cycles.)

## Impact

Any template where a resource is declared before something it references — very
common with SAM state machines that substitute a Lambda's name into their
definition — deploys "successfully" but is broken at runtime
(`Lambda.ResourceNotFoundException`). It's deterministic, not order-of-luck, once
the declaration order trips it.

Happy to test a branch or send a PR.

## The fix

Add a `dependency_order` topological sort in
`crates/fakecloud-cloudformation/src/template/resolution.rs` and wire it into the
changeset provisioning loop (`service.rs`), so a referenced resource is provisioned
**before** its referrer and `resolve_refs_full` returns the physical name/ARN
instead of the logical-id placeholder. Edges are drawn from explicit `DependsOn`
plus implicit `Ref` / `Fn::GetAtt` / `Fn::Sub` references (collected recursively,
honoring the `${!X}` literal escape in `Fn::Sub`).

## Design notes / alternatives considered

- **Topological provisioning order vs. a post-provision placeholder rewrite.**
  Ordering is the AWS-correct model: real CloudFormation provisions in dependency
  order, so a `Ref` always resolves to a created resource. It fixes the whole class
  of Ref-resolution bugs at the source, not just the state-machine
  `DefinitionSubstitutions` case. A post-hoc rewrite (re-scan provisioned resources
  and patch baked-in logical ids) is narrower and more fragile — it has to know
  every place a logical id might have been baked (ASL, env vars, inline policies, …)
  and re-serialize each, and still races anything that already read the wrong value.
- **Cycle handling.** The sort is Kahn's algorithm; if a cycle leaves resources
  unresolved, the remainder is appended in original declaration order (graceful
  degradation) rather than panicking. Real CloudFormation rejects dependency cycles
  outright; this never makes a cyclic template worse than today's behavior.
- **Deterministic ordering for equal-priority resources.** The ready set is ordered
  by original template index (and the graph is built with ordered `BTree*`
  collections throughout), so resources with no dependency relationship keep their
  declared order — the sort is stable and reproducible, not hash-ordered.
- **Edge sources.** `DependsOn` (explicit) plus implicit `Ref`, `Fn::GetAtt`, and
  `Fn::Sub` references, collected recursively through nested template values.

## Test plan

- `cargo test -p fakecloud-cloudformation` — `dependency_order` unit tests +
  `changeset_provisions_lambda_before_referencing_state_machine` (a StateMachine
  declared/sorting before the Lambda it references resolves to the function name,
  not the logical id, in the baked ASL).
- `cargo test -p fakecloud-e2e --test cloudformation --test cloudformation_stepfunctions --test cloudformation_sam_statemachine` — green.

---
## Summary by cubic
Provision CloudFormation resources in dependency order so references resolve to physical IDs instead of logical IDs. This matches real CFN and fixes runtime failures (e.g., Step Functions → Lambda substitutions).

- **Bug Fixes**
  - Topologically sort resources from `DependsOn`, `Ref`, `Fn::GetAtt`, and `Fn::Sub` (respects `${!}` escapes, ignores local `Fn::Sub` vars, built after `ForEach`/SAM expansion); deterministic tie-breaking, cycles fall back to template order.
  - Apply `template::dependency_order` in both `CreateStack` and change set paths (`provision_stack_resources`, `apply_resource_updates`); adds tests including the StateMachine→Lambda case and updates ECS e2e so a service `Ref`s the re-registered task definition revision (`cfn-task:2`) after update.

<sup>Written for commit f3e2a27ca96e465f9d7f0e3857a5113f1d6dfab8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/faiscadev/fakecloud/pull/1719?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/fakecloud-cloudformation/src/extras.rs
+++ b/crates/fakecloud-cloudformation/src/extras.rs
@@ -2842,6 +2842,87 @@ mod tests {
         assert!(!stack.outputs[0].value.is_empty());
     }
 
+    // Gap #3: a StateMachine that references a Lambda via DefinitionSubstitutions
+    // must provision *after* the Lambda even when it sorts/declares first, so the
+    // substitution resolves to the function name rather than leaving the logical
+    // id baked in the ASL (which broke every invoke with Lambda.ResourceNotFound).
+    #[test]
+    fn changeset_provisions_lambda_before_referencing_state_machine() {
+        let d = deps();
+        let sfn = d.stepfunctions.clone();
+        let state: SharedCloudFormationState =
+            Arc::new(RwLock::new(MultiAccountState::<CloudFormationState>::new(
+                "000000000000",
+                "us-east-1",
+                "",
+            )));
+        let svc = CloudFormationService::new(state, d);
+
+        // "Machine" sorts before "Worker", so without dependency ordering the
+        // StateMachine provisions first and bakes the unresolved logical id.
+        let template = r#"{
+            "Resources": {
+                "Machine": {
+                    "Type": "AWS::StepFunctions::StateMachine",
+                    "Properties": {
+                        "RoleArn": "arn:aws:iam::000000000000:role/sfn",
+                        "DefinitionString": "{\"StartAt\":\"T\",\"States\":{\"T\":{\"Type\":\"Task\",\"Resource\":\"${fn}\",\"End\":true}}}",
+                        "DefinitionSubstitutions": {"fn": {"Ref": "Worker"}}
+                    }
+                },
+                "Worker": {
+                    "Type": "AWS::Lambda::Function",
+                    "Properties": {
+                        "FunctionName": "workflow_dispatcher_v2-1",
+                        "Runtime": "python3.12",
+                        "Handler": "index.handler",
+                        "Role": "arn:aws:iam::000000000000:role/lambda",
+                        "Code": {"ZipFile": "def handler(e, c): return e"}
+                    }
+                }
+            }
+        }"#;
+
+        svc.handle_extra_action(&req(
+            "CreateChangeSet",
+            &[
+                ("StackName", "wf"),
+                ("ChangeSetName", "cs1"),
+                ("ChangeSetType", "CREATE"),
+                ("TemplateBody", template),
+            ],
+        ))
+        .expect("create change set");
+        svc.handle_extra_action(&req(
+            "ExecuteChangeSet",
+            &[("StackName", "wf"), ("ChangeSetName", "cs1")],
+        ))
+        .expect("execute change set");
+
+        let accounts = sfn.read();
+        let st = accounts.get("000000000000").expect("sfn account exists");
+        let machine = st
+            .state_machines
+            .values()
+            .next()
+            .expect("state machine was provisioned");
+        assert!(
+            machine.definition.contains("workflow_dispatcher_v2-1"),
+            "ASL should carry the resolved function name, got: {}",
+            machine.definition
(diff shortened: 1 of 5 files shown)


WarpRat:
Tracking issue: #1720

vieiralucas:
Thanks @WarpRat! The topological dependency ordering is the right fix for the baked-logical-id problem (`Lambda.ResourceNotFoundException`). Heads up on one thing I adjusted: your change correctly provisions a referenced resource before its dependent, which surfaced a latent issue in the existing ECS e2e — after an UpdateStack that changes a task definition, the service now resolves `Ref: TaskDef` to the re-registered revision (`cfn-task:2`), matching real CloudFormation. The old assertion (`cfn-task:1`) only passed because of the stale ordering your PR fixes, so I updated it (verified locally). Also fixed one unicode arrow. Merged.

PR #1733, opened 2026-06-15
feat(lambda): configurable k8s Pod nodeSelector, tolerations, and annotations

WarpRat:
Thanks again, @vieiralucas, for your quick review and release on all the previous updates to the k8s backend for lambdas. These are the last minor improvements that would help this fit perfectly into our pipelines. We currently segregate feature environments and testing CI runners onto node pools that allow spot nodes. By providing a mechanism to set tolerations and node selectors, we can make sure the lambda pods land on the same nodes as the rest of the feature env stack. I don't currently have a need to set any of these options per lambda, but I could see situations where other users may find that helpful.

I haven't had a chance to test out the k8s backend for any of the other services, but I plan to experiment with those sometime in the near future and will certainly let you know if we hit any issues with them. Since this change makes some design decisions and isn't strictly necessary for us to use fakecloud, I wanted to get your input or input from anyone else in the community who may have opinions. If the environment variables and tag overrides sound good, I can open the other MRs immediately.

## Summary

Adds operator-configurable Pod **node selectors**, **taint tolerations**, and **arbitrary annotations** to the Kubernetes backend, starting with **Lambda**. Each of the three is resolvable at three levels, lowest to highest precedence:

1. **Global** env: `FAKECLOUD_K8S_NODE_SELECTOR` / `FAKECLOUD_K8S_TOLERATIONS` / `FAKECLOUD_K8S_ANNOTATIONS` (apply to every fakecloud Pod)
2. **Per-service** env: `FAKECLOUD_LAMBDA_K8S_NODE_SELECTOR` / `_TOLERATIONS` / `_ANNOTATIONS`
3. **Per-function** reserved tags: `fakecloud-k8s/node-selector`, `fakecloud-k8s/tolerations`, `fakecloud-k8s/annotations`

The shared parsing/merge/apply logic lives in a new `fakecloud-k8s` `pod_config` module (`K8sPodConfig`) so the remaining k8s-backed services can reuse it unchanged.

## Motivation

When fakecloud runs the k8s backend in a real cluster, operators routinely need to steer where the spawned Pods land and how they're annotated: pin them to a node pool (`nodeSelector`), let them run on tainted/spot nodes (`tolerations`), or attach metadata for cost allocation / mesh sidecar injection / scrape config (`annotations`). Today the backend sets none of these, so there's no way to influence scheduling of the Pods fakecloud creates. This adds that control without changing any default behavior (unset = no-op).

### Configuration model and precedence

`global env  ->  service env  ->  per-instance tag`, applied lowest-to-highest. A backend reads the global+service base once at startup; the per-function tags are merged over it at launch.

- **Maps** (`node_selector`, `annotations`): union per key; a higher level overrides only the keys it sets, lower-level keys survive.
- **Tolerations** (list): combine **additively** across levels (a Pod tolerates the union of taints), dropping exact duplicates. This mirrors the repo's existing SAM Globals convention (maps deep-merge, list-typed keys combine additively).

### Value formats

- `NODE_SELECTOR` / `ANNOTATIONS`: a flat `key=value,key=value` map.
- `TOLERATIONS`: a JSON array of Kubernetes `Toleration` objects, parsed via `serde_json` straight into `k8s_openapi`'s `Toleration` type (full field fidelity, no hand-rolled list parser).

### Error handling (deliberately asymmetric)

- A malformed **env** value (e.g. invalid tolerations JSON) **fails fast** at fakecloud startup, so an operator typo is loud.
- A malformed per-instance **tag** is **logged and ignored for that field**, so a bad tag can never make a single function un-runnable.

### Example

```sh
# Global: keep every fakecloud Pod off the default node pool.
FAKECLOUD_K8S_NODE_SELECTOR=fakecloud=true
FAKECLOUD_K8S_TOLERATIONS='[{"key":"fakecloud","operator":"Exists","effect":"NoSchedule"}]'

# Service: send all Lambda Pods to the burst pool, with a cost-center annotation.
FAKECLOUD_LAMBDA_K8S_NODE_SELECTOR=pool=burst
FAKECLOUD_LAMBDA_K8S_ANNOTATIONS=cost-center=lambda
```

```sh
# Per-function: this one function pins to GPU nodes (overrides the service/global
# nodeSelector for that key) via a tag at create time.
aws lambda create-function \
  --tags 'fakecloud-k8s/node-selector=accelerator=nvidia,fakecloud-k8s/annotations=team=ml' \
  ...
```

## Scope / non-goals

- **Resource requests/limits are intentionally untouched.** Lambda already derives memory `requests`==`limits` from the function's `MemorySize`; this PR does not change that.
- **Affinity / anti-affinity is intentionally excluded.** It maps to no real AWS Lambda behavior, and its semantics vary widely across schedulers/autoscalers, which makes it an easy footgun for someone just getting started with the k8s backend. Easy to add later behind the same `apply` pattern if there's demand.

## Implementation notes

- New `crates/fakecloud-k8s/src/pod_config.rs`: `K8sPodConfig { node_selector, tolerations, annotations }` plus `from_env_prefix`, `resolved_base(service_prefix)`, `from_tags`, `merge`, and `apply(&mut Pod)`. Re-exported from the crate root.
- Lambda backend stores the global+service base at `from_env` (`K8sPodConfig::resolved_base("FAKECLOUD_LAMBDA_K8S")`) and, at launch, merges the function's tags over it and applies the result to the built Pod before `create_pod`.
- `apply` mutates the already-built Pod, so the per-service Pod builders don't change signature -- which keeps the follow-up service PRs small.

## Tests

- 22 unit tests for `pod_config` covering: `key=value` map parsing edge cases (whitespace, trailing comma, missing `=`, empty key, duplicate key, value containing `=`); valid/invalid tolerations JSON; `merge` precedence (maps per-key override, tolerations additive + dedupe); `from_tags` warn-and-skip on malformed input; `apply` (each field set/unset, annotation union with pre-existing, empty no-op, `spec: None`); and env parsing incl. global+service resolution and fail-fast on bad tolerations.
- A Lambda test asserting a per-function tag overrides the global/service base and reaches the built Pod.
- All existing `fakecloud-lambda` k8s unit tests still pass; `cargo clippy` and `cargo fmt --check` clean on both crates; full workspace test suite (excluding the docker-dependent e2e crate) green.

## Docs

- New "Pod scheduling and metadata" section in `website/content/docs/guides/kubernetes-backend.md` (env vars, tag keys, formats, precedence, merge rules, and the create-time-tag caveat).
- New rows in `website/content/docs/reference/configuration.md`.

---

## Coming next: the same for the other four k8s backends

We have the same change drafted for **RDS, ElastiCache, ECS, and EC2**, all built on the shared `pod_config` module and the identical `global -> service -> per-instance-tag` model (`FAKECLOUD_RDS_K8S_*`, `FAKECLOUD_ELASTICACHE_K8S_*`, `FAKECLOUD_ECS_K8S_*`, `FAKECLOUD_EC2_K8S_*`, plus the same `fakecloud-k8s/*` tags). We'd like your input on one design point before we open them, so we land them the way you'd prefer rather than asking you to redo it.

For those four services, the only extra work versus Lambda is wiring each Pod-builder to read the resource's tags (Lambda already had the function in scope at build time; the others currently receive only an identifier). That plumbing is uniform and needed regardless.

### The one open question: RDS create-time tags

While drafting these we found a relevant asymmetry in how create-time `Tags` are persisted today:

| Service | Persists tags from the **create** call? | Persists later `AddTags`/`TagResource`? |
| --- | --- | --- |
| Lambda | yes | yes |
| ElastiCache | yes | yes |
| ECS (`RunTask`) | yes (incl. `propagateTags`) | yes |
| EC2 (`RunInstances`) | yes (`TagSpecifications`) | yes |
| **RDS (`CreateDBInstance`)** | **no -- request `Tags` are dropped** | yes |

So for ElastiCache/ECS/EC2, per-instance scheduling will work at create time as soon as we wire the builder. For **RDS**, because `CreateDBInstance` discards the request's `Tags` today, per-instance scheduling at create won't work unless we also make `CreateDBInstance` persist its `Tags`. Two ways to go:

- **(A) Also fix RDS `CreateDBInstance` to persist `Tags`.** Small, isolated, and arguably just more AWS-correct (real RDS applies create-time tags). Per-instance scheduling then works uniformly at create across all five services. We'd run the full workspace suite to confirm no regression, and could split the tag-persistence fix into its own PR if you'd rather review it separately.
- **(B) Leave RDS tag persistence as-is.** Only wire the builder to read whatever tags are in state. RDS per-instance scheduling then applies only after `AddTagsToResource` (+ a reboot to recreate the Pod), consistent with the "recreate to pick up tags" caveat in the docs. The other four still work at create.

We lean toward (A), but it's your call.

### A couple of smaller choices we'd value your opinion on

- **Tag key namespace:** we used `fakecloud-k8s/node-selector` (etc.). Happy to change the prefix/spelling if you have a convention you prefer.
- **Tolerations merge:** we chose additive-union across the three levels (to match the SAM Globals list-merge convention). If you'd rather a higher level fully *replace* lower-level tolerations, that's a one-line change.
- **Per-instance tags at create vs. recreate:** the current behavior is that tags added after creation (via `AddTags`) only affect the Pod on the next recreate/reboot, which we documented as a caveat. Shout if you'd prefer different semantics.

Happy to open the four follow-up PRs immediately once you've weighed in on the above (especially the RDS question).

---
## Summary by cubic
Adds operator-configurable Kubernetes Pod scheduling and metadata for Lambda: set `nodeSelector`, `tolerations`, and `annotations` via env and per-function tags to steer where Pods run. Defaults stay the same; only applies when set. Docs clarify that `fakecloud-k8s/*` tags remain visible and count toward tag limits.

- **New Features**
  - Configure Pod `nodeSelector`, `tolerations`, and `annotations` for k8s Lambda Pods.
  - Precedence: global `FAKECLOUD_K8S_*` → service `FAKECLOUD_LAMBDA_K8S_*` → per-function tags (`fakecloud-k8s/node-selector`, `fakecloud-k8s/tolerations`, `fakecloud-k8s/annotations`).
  - Merge: maps union per key; tolerations additive with dedupe. Formats: `key=value,key=value` for maps; tolerations as a JSON array of Kubernetes `Toleration`. Unset = no-op.
  - Robustness: bad env fails at startup; bad per-function tag is logged and ignored.
  - Shared logic in `fakecloud-k8s` as `K8sPodConfig` for reuse. Docs note `fakecloud-k8s/*` tags show up in tag listings and count toward AWS tag limits.

<sup>Written for commit 0fb6c0032aa368eff1fc5a92bac44492754c5ed2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/faiscadev/fakecloud/pull/1733?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/fakecloud-k8s/src/lib.rs
+++ b/crates/fakecloud-k8s/src/lib.rs
@@ -13,8 +13,10 @@
 //! Backend selection (`FAKECLOUD_CONTAINER_BACKEND` /
 //! `FAKECLOUD_<SERVICE>_BACKEND`) lives in [`backend`]. The kube client
 //! wrapper and Pod lifecycle live in [`client`]. Env parsing
-//! (`FAKECLOUD_K8S_*`) lives in [`env`]. Naming + label conventions live
-//! in [`names`] and [`labels`].
+//! (`FAKECLOUD_K8S_*`) lives in [`env`]. Operator-configurable Pod
+//! scheduling/metadata (node selector, tolerations, annotations) lives in
+//! [`pod_config`]. Naming + label conventions live in [`names`] and
+//! [`labels`].
 //!
 //! See `website/content/docs/guides/kubernetes-backend.md` for the
 //! operator-facing setup (ServiceAccount, RBAC, Deployment yaml).
@@ -24,7 +26,9 @@ pub mod client;
 pub mod env;
 pub mod labels;
 pub mod names;
+pub mod pod_config;
 
 pub use backend::{backend_choice, Backend};
 pub use client::{ExecOutput, K8sClient, K8sError};
 pub use env::{K8sEnv, K8sEnvError};
+pub use pod_config::{K8sPodConfig, K8sPodConfigError};
--- a/crates/fakecloud-k8s/src/pod_config.rs
+++ b/crates/fakecloud-k8s/src/pod_config.rs
@@ -1,0 +1,552 @@
+//! Operator-configurable Pod scheduling + metadata shared by every
+//! FakeCloud k8s backend.
+//!
+//! Lets operators attach a `nodeSelector`, taint `tolerations`, and
+//! arbitrary `annotations` to the Pods FakeCloud creates, without each
+//! service re-implementing the parsing and application. Each of the three
+//! is resolvable at three levels, lowest to highest precedence:
+//!
+//! 1. Global env  - `FAKECLOUD_K8S_<KEY>`
+//! 2. Service env - `FAKECLOUD_<SERVICE>_K8S_<KEY>`
+//!    (e.g. `FAKECLOUD_LAMBDA_K8S_NODE_SELECTOR`)
+//! 3. Per-instance tag - a reserved tag on the resource itself
+//!    (e.g. a Lambda function tag `fakecloud-k8s/node-selector`)
+//!
+//! A backend reads the global+service base once at startup with
+//! [`K8sPodConfig::resolved_base`], then per-Pod merges the resource's
+//! tags over it with [`K8sPodConfig::merge`] + [`K8sPodConfig::from_tags`]
+//! and writes the result onto the built Pod with [`K8sPodConfig::apply`].
+//!
+//! Value formats:
+//! - `NODE_SELECTOR` / `ANNOTATIONS`: a flat `key=value,key=value` map.
+//! - `TOLERATIONS`: a JSON array of k8s `Toleration` objects.
+//!
+//! Merge semantics across the three levels: the maps (node selector,
+//! annotations) union per key, with the higher level winning on a key
+//! clash; tolerations combine additively (a Pod tolerates the union of
+//! taints), dropping exact duplicates. This mirrors the repo's SAM
+//! Globals convention (maps deep-merge, list-typed keys combine).
+//!
+//! Error handling is asymmetric on purpose: a malformed *env* value fails
+//! fast at startup (an operator typo should be loud), but a malformed
+//! per-instance *tag* is warned about and ignored for that field -- a bad
+//! tag must never make a single resource un-runnable.
+
+use std::collections::BTreeMap;
+
+use k8s_openapi::api::core::v1::{Pod, Toleration};
+
+/// Reserved per-instance tag key carrying a `key=value,key=value`
+/// node-selector for this resource's Pod.
+pub const TAG_NODE_SELECTOR: &str = "fakecloud-k8s/node-selector";
+/// Reserved per-instance tag key carrying a JSON array of `Toleration`s.
+pub const TAG_TOLERATIONS: &str = "fakecloud-k8s/tolerations";
+/// Reserved per-instance tag key carrying `key=value,key=value`
+/// annotations merged onto this resource's Pod.
+pub const TAG_ANNOTATIONS: &str = "fakecloud-k8s/annotations";
+
+/// The global env prefix every service's base config inherits from.
+const GLOBAL_PREFIX: &str = "FAKECLOUD_K8S";
+
+/// Resolved node selector, tolerations, and annotations for a Pod.
+#[derive(Debug, Clone, Default, PartialEq)]
+pub struct K8sPodConfig {
+    /// `spec.nodeSelector` entries.
+    pub node_selector: BTreeMap<String, String>,
+    /// `spec.tolerations` entries.
+    pub tolerations: Vec<Toleration>,
+    /// `metadata.annotations` entries.
+    pub annotations: BTreeMap<String, String>,
+}
+
+/// Errors parsing the `FAKECLOUD_*_K8S_*` Pod configuration. Only env
+/// parsing fails fast; per-instance tags warn and fall back instead.
+#[derive(Debug, thiserror::Error)]
+pub enum K8sPodConfigError {
+    #[error("{0} is not a valid JSON array of Tolerations: {1}")]
+    InvalidTolerations(String, String),
+}
+
+impl K8sPodConfig {
+    /// Read a single env level. `prefix` is the part before the key, e.g.
+    /// `FAKECLOUD_K8S` (global) or `FAKECLOUD_LAMBDA_K8S` (service). Reads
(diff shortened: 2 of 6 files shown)


vieiralucas:
Merged, thanks @WarpRat, this is a clean, well-tested piece of work. The shared `K8sPodConfig` module with the `apply`-onto-built-Pod pattern is exactly the right shape for fanning this out to the other backends, and the asymmetric error handling (env fails fast, per-instance tag warns-and-skips) is the correct call.

I pushed one tiny doc follow-up before merging: a note that the `fakecloud-k8s/*` tags stay visible in `ListTags`/`list-tags-for-resource` and count toward the resource tag limit, since fakecloud reads but does not strip them. That is the AWS-correct behavior (the user set them), just worth making explicit.

On your open questions, green light across the board:

- **RDS create-time tags, option (A).** Yes, fix `CreateDBInstance` to persist its request `Tags`. Real RDS applies create-time tags, so dropping them today is a genuine correctness gap, not just a blocker for scheduling. Please **split it into its own PR** as you offered. It is a behavior fix and deserves its own review/changelog line, separate from the scheduling wiring. Once it lands, per-instance scheduling works uniformly at create across all five services.
- **Tag namespace `fakecloud-k8s/*`:** keep it, no change.
- **Tolerations additive-union merge:** keep it. Matching the SAM Globals list-merge convention is the right consistency choice.
- **Create-vs-recreate tag semantics:** keep the documented "recreate to pick up later-added tags" behavior. It is the honest model and well-captured in the guide.

So: please go ahead and open the four follow-up PRs (RDS / ElastiCache / ECS / EC2) on the shared `pod_config` module, with the RDS tag-persistence fix as its own separate PR. Looking forward to them.

PR #1741, opened 2026-06-15
feat(rds): configurable k8s Pod nodeSelector, tolerations, and annotations

WarpRat:
## Summary

Wires the shared `K8sPodConfig` module (added in #1733) into the **RDS** k8s backend, so DB instance Pods honor operator-configured node selectors, tolerations, and annotations. Second of the four follow-up PRs you green-lit on #1733.

Resolution is the same three-level model as Lambda, lowest to highest precedence:

1. **Global** env: `FAKECLOUD_K8S_NODE_SELECTOR` / `_TOLERATIONS` / `_ANNOTATIONS`
2. **Per-service** env: `FAKECLOUD_RDS_K8S_NODE_SELECTOR` / `_TOLERATIONS` / `_ANNOTATIONS`
3. **Per-instance** reserved tags on the DB instance: `fakecloud-k8s/node-selector`, `fakecloud-k8s/tolerations`, `fakecloud-k8s/annotations`

Maps merge per key (higher wins); tolerations combine additively; malformed env fails fast, malformed tag warns-and-skips — all inherited from the shared module.

## Implementation

- `K8sDb` stores the global+service base (`K8sPodConfig::resolved_base("FAKECLOUD_RDS_K8S")`) at `from_env`.
- `ensure` merges the instance's reserved-tag overrides over the base and `apply`s the result to the built Pod **before it is cached**, so a reboot (which recreates from the cached spec) reuses the same scheduling.
- The DB instance's tags are plumbed to the builder through `ensure_postgres` at every launch site: create, start, crash-recovery, read-replica, restore-from-snapshot, restore-to-point-in-time, and restore-from-S3. A **read replica inherits its source instance's** scheduling (there's no separate replica tag input on that path).

## Relationship to the other PRs

- Builds on the shared `pod_config` module from #1733.
- Independent of the RDS create-time tag fix (#1740), but complementary: until #1740 lands, `CreateDBInstance` stores no tags, so per-instance scheduling on a brand-new instance takes effect after `AddTagsToResource` + a reboot. With #1740, it works at create time too. This PR needs no changes either way.

## Tests

- Unit test `pod_config_overrides_apply_to_built_pod` builds a real DB Pod via `build_pod` and asserts the merged base+tag config (node selector override + added annotation) lands on it — mirroring the `ensure` wiring.
- Full `fakecloud-rds` unit suite green; `cargo clippy` / `cargo fmt --check` clean. The k8s integration suite is feature-gated and needs a cluster (not run here).

## Docs

The general "Pod scheduling and metadata" guide section from #1733 already documents the model and the `FAKECLOUD_<SERVICE>_K8S_*` / tag conventions for all services, so no per-service doc change is needed here.

---
## Summary by cubic
Adds configurable Kubernetes Pod scheduling for RDS. DB Pods now honor nodeSelector, tolerations, and annotations with a clear precedence and consistent reuse across reboots.

- **New Features**
  - Integrates shared `K8sPodConfig` into the RDS k8s backend.
  - Precedence: global `FAKECLOUD_K8S_*`, service `FAKECLOUD_RDS_K8S_*`, then per-instance tags `fakecloud-k8s/node-selector`, `fakecloud-k8s/tolerations`, `fakecloud-k8s/annotations`; per-key merge, tolerations additive; malformed env fails fast, malformed tags warn-and-skip.
  - Applies config before the Pod spec is cached, so reboots keep the same scheduling.
  - Plumbs instance tags through all launch paths (create/start/crash-recovery/read-replica/restores). Read replicas inherit the source instance’s scheduling.
  - Adds a unit test asserting the merged config lands on the built Pod.

<sup>Written for commit 5ce95bea89ec84706ea539351557bfbdb82b5bcb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/faiscadev/fakecloud/pull/1741?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/fakecloud-rds/src/runtime/k8s.rs
+++ b/crates/fakecloud-rds/src/runtime/k8s.rs
@@ -29,7 +29,7 @@ use k8s_openapi::apimachinery::pkg::apis::meta::v1::ObjectMeta;
 use parking_lot::RwLock;
 use tokio_postgres::NoTls;
 
-use fakecloud_k8s::{labels, names, K8sClient, K8sEnv};
+use fakecloud_k8s::{labels, names, K8sClient, K8sEnv, K8sPodConfig};
 
 use super::{bridge_image_tag, BackendInitError, RunningDbContainer, RuntimeError};
 
@@ -67,6 +67,10 @@ pub(super) struct K8sDb {
     /// fakecloud.
     self_url: String,
     pull_secret: Option<String>,
+    /// Global + RDS-service node selector / tolerations / annotations
+    /// applied to every DB Pod. Per-instance tag overrides are merged over
+    /// this when the Pod is built.
+    pod_config: K8sPodConfig,
     /// Built Pod spec + port per instance id, so a reboot can recreate the
     /// Pod without re-deriving the engine config.
     specs: Arc<RwLock<HashMap<String, (Pod, u16)>>>,
@@ -84,6 +88,7 @@ impl std::fmt::Debug for K8sDb {
 impl K8sDb {
     pub(super) async fn from_env(server_port: u16) -> Result<Self, BackendInitError> {
         let env = K8sEnv::from_env(server_port)?;
+        let pod_config = K8sPodConfig::resolved_base("FAKECLOUD_RDS_K8S")?;
         let client = K8sClient::connect(env.namespace.clone())
             .await
             .map_err(|e| BackendInitError::Connect(e.to_string()))?;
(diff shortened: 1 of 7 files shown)


vieiralucas:
Merged, thanks @WarpRat! 🎉 Threading the reserved tags through every `ensure_postgres` call site (create, reboot, replica, restore, snapshot-restore) was exactly right, and inheriting the source instance's scheduling for read replicas is a sensible default. Builds on #1740 cleanly.

PR #1742, opened 2026-06-15
feat(elasticache): configurable k8s Pod nodeSelector, tolerations, and annotations

WarpRat:
## Summary

Wires the shared `K8sPodConfig` module (#1733) into the **ElastiCache** k8s backend, so cache Pods honor operator-configured node selectors, tolerations, and annotations. Third of the four follow-ups you green-lit.

- **Global** env: `FAKECLOUD_K8S_NODE_SELECTOR` / `_TOLERATIONS` / `_ANNOTATIONS`
- **Per-service** env: `FAKECLOUD_ELASTICACHE_K8S_NODE_SELECTOR` / `_TOLERATIONS` / `_ANNOTATIONS`

The service base (global merged with the ElastiCache-service overrides) is resolved once at `from_env` and applied in `spawn_pod_bytes` — the single chokepoint that **every** spawn path funnels through (CreateCacheCluster, serverless, replication groups, crash-recovery, and RebootCacheCluster), so all of them are covered uniformly with no per-call-site changes.

## Per-instance tags: deferred (scoping note)

Lambda (#1733) and RDS (#1741) also support per-instance overrides via `fakecloud-k8s/*` tags, because their tags live on the resource struct and are in scope at Pod-build time. ElastiCache stores tags in a separate by-ARN map (`ElastiCacheState.tags`), so threading per-instance tags down to the builder is a larger, separate change across the create/serverless/replication/recovery/reboot paths. This PR delivers the global + per-service levels (which cover the cluster-wide scheduling use case); per-instance tag overrides for ElastiCache can follow using the same `K8sPodConfig::from_tags` + `merge` helpers. Happy to do that follow-up if you'd like it in this PR instead.

## Tests

- Unit test asserting the resolved config applies (node selector + annotation) to a built cache Pod.
- Full `fakecloud-elasticache` unit suite green; `cargo clippy` / `cargo fmt --check` clean. K8s integration suite is feature-gated (needs a cluster); not run here.

## Docs

The general "Pod scheduling and metadata" section added in #1733 already covers the env-var model and the `FAKECLOUD_<SERVICE>_K8S_*` convention for all services.

---
## Summary by cubic
Adds configurable scheduling and metadata for ElastiCache k8s Pods via `K8sPodConfig`. Pods now honor global and per-service env defaults and per-instance `fakecloud-k8s/*` tag overrides across all spawn paths, including reboot and recovery.

- **New Features**
  - Builds a base from `FAKECLOUD_K8S_*` and `FAKECLOUD_ELASTICACHE_K8S_*`.
  - Merges per-instance `fakecloud-k8s/node-selector`, `fakecloud-k8s/tolerations`, and `fakecloud-k8s/annotations` over the base (per-instance wins).
  - Applies config in the spawn chokepoint for create, serverless, replication, recovery, and reboot; reboot/recovery re-derive tags from state to keep node placement.

- **Migration**
  - To control scheduling/annotations, set `FAKECLOUD_K8S_*` or `FAKECLOUD_ELASTICACHE_K8S_*`, and optionally add `fakecloud-k8s/*` tags on resources.
  - No other changes; Docker backend ignores these settings.

<sup>Written for commit 58fdf04f6b982686d8e3d470e6d3c3d0eb3971f2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/faiscadev/fakecloud/pull/1742?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/fakecloud-elasticache/src/runtime/k8s.rs
+++ b/crates/fakecloud-elasticache/src/runtime/k8s.rs
@@ -101,6 +101,7 @@ impl K8sCache {
         resource_id: &str,
         engine: CacheEngineKind,
         rdb_path: Option<&str>,
+        tags: &std::collections::BTreeMap<String, String>,
     ) -> Result<RunningCacheContainer, RuntimeError> {
         let rdb = match rdb_path {
             Some(path) => Some(tokio::fs::read(path).await.map_err(|e| {
@@ -108,7 +109,7 @@ impl K8sCache {
             })?),
             None => None,
         };
-        self.spawn_pod_bytes(resource_id, engine, rdb).await
+        self.spawn_pod_bytes(resource_id, engine, rdb, tags).await
     }
 
     async fn spawn_pod_bytes(
@@ -116,6 +117,7 @@ impl K8sCache {
         resource_id: &str,
         engine: CacheEngineKind,
         rdb: Option<Vec<u8>>,
+        tags: &std::collections::BTreeMap<String, String>,
     ) -> Result<RunningCacheContainer, RuntimeError> {
         let pod_name = names::pod_name(POD_PREFIX, resource_id, resource_id);
         let port = engine.port();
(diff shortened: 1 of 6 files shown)


vieiralucas:
Merged, thanks! 🎉

One adjustment I pushed directly to this branch before merging: the original revision wired up only the **global + service-env** layer (`FAKECLOUD_ELASTICACHE_K8S_*`) and applied it base-only. The k8s-backend docs (`kubernetes-backend.md`, the precedence list landed in #1733) already promise per-instance reserved `fakecloud-k8s/*` tags on **cache clusters** specifically, so base-only would have been a silent footgun where the same tag works on RDS/ECS/Lambda but gets ignored here.

So I threaded the per-instance layer through to match the documented 3-level precedence (global, then service, then per-resource tag), covering every spawn path:
- `CreateCacheCluster` / `CreateReplicationGroup` / `CreateServerlessCache` read the create-time `Tags`
- `RebootCacheCluster` and the post-restart recovery re-derive the reserved tags from persisted state, so a Pod recreate keeps its node placement

Same treatment went into the EC2 PR (#1744). Added an override unit test alongside your base-apply one. All green, merged.

PR #1744, opened 2026-06-15
feat(ec2): configurable k8s Pod nodeSelector, tolerations, and annotations

WarpRat:
## Summary

Wires the shared `K8sPodConfig` module (#1733) into the **EC2** k8s backend, so instance Pods honor operator-configured node selectors, tolerations, and annotations. Last of the four follow-ups you green-lit.

- **Global** env: `FAKECLOUD_K8S_NODE_SELECTOR` / `_TOLERATIONS` / `_ANNOTATIONS`
- **Per-service** env: `FAKECLOUD_EC2_K8S_NODE_SELECTOR` / `_TOLERATIONS` / `_ANNOTATIONS`

The service base (global merged with the EC2-service overrides) is resolved once at `from_env` and applied in `spawn_pod` — the single chokepoint every spawn goes through (`RunInstances`, and the `Start`/`Reboot` recreate path) — so all instance Pods get the configured scheduling/metadata uniformly.

## Per-instance tags: deferred (scoping note)

Like ElastiCache (#1742), EC2 stores tags in a separate by-ARN map (`Ec2State.tags`) rather than on the `Instance` struct, so threading per-instance `fakecloud-k8s/*` tag overrides down to the builder is a larger, separate change. This PR delivers the global + per-service levels (which cover the cluster-wide scheduling use case — pinning instances to a node pool / tolerating spot taints). Per-instance overrides can follow using the same `K8sPodConfig::from_tags` + `merge` helpers; happy to fold that in if you'd prefer.

(For contrast: Lambda, RDS, and ECS keep tags on the resource struct, so those PRs wired per-instance overrides too.)

## Tests

- Unit test asserting the resolved config applies (node selector + annotation) to a built instance Pod.
- Full `fakecloud-ec2` unit suite green; `cargo clippy` / `cargo fmt --check` clean. K8s integration suite is feature-gated (needs a cluster); not run here.

## Docs

The "Pod scheduling and metadata" section from #1733 already documents the env-var model and the `FAKECLOUD_<SERVICE>_K8S_*` convention for all services.

---
## Summary by cubic
Adds configurable Kubernetes Pod scheduling and metadata to the EC2 backend. Instance Pods now honor node selectors, tolerations, and annotations from env vars and per-instance `fakecloud-k8s/*` tags on every spawn.

- **New Features**
  - Merges `K8sPodConfig` from global `FAKECLOUD_K8S_*`, per-service `FAKECLOUD_EC2_K8S_*`, and per-instance `fakecloud-k8s/*` tag overrides.
  - Resolves the base at init and applies it in `spawn_pod` for `RunInstances`, `Start`, and `Reboot`.
  - Parses overrides from `RunInstances` `TagSpecification` and stores them on the instance so recreates reuse the same scheduling.
  - Extends `BackendInitError` with `K8sPodConfigError` for invalid config.

<sup>Written for commit 46809ea4c1064f3fa4ffd2178fe33277aefaf00c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/faiscadev/fakecloud/pull/1744?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/fakecloud-ec2/src/runtime/k8s.rs
+++ b/crates/fakecloud-ec2/src/runtime/k8s.rs
@@ -82,6 +82,7 @@ impl K8sInstances {
         instance_id: &str,
         image: &str,
         user_data: Option<&str>,
+        tags: &std::collections::BTreeMap<String, String>,
     ) -> Result<RunningInstance, RuntimeError> {
         let seq = self.spawn_seq.fetch_add(1, Ordering::Relaxed);
         let pod_name = names::pod_name(POD_PREFIX, instance_id, &format!("{instance_id}-{seq}"));
@@ -94,10 +95,14 @@ impl K8sInstances {
             user_data,
             pull_secret: self.pull_secret.as_deref(),
         });
-        // Operator-configured global + EC2-service scheduling / metadata.
-        // Every spawn (RunInstances, and Start/Reboot recreate) goes
-        // through here, so this covers them all.
-        self.pod_config.apply(&mut pod);
+        // Operator-configured global + EC2-service base, with this
+        // instance's reserved `fakecloud-k8s/*` tag overrides merged over it
+        // (per-instance wins). Every spawn (RunInstances, and Start/Reboot
+        // recreate) goes through here, so this covers them all.
+        self.pod_config
+            .clone()
+            .merge(K8sPodConfig::from_tags(tags))
+            .apply(&mut pod);
 
         self.client
             .create_pod(&pod)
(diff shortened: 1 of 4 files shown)


vieiralucas:
Merged, thanks! 🎉

Same adjustment as the ElastiCache PR (#1742): pushed a commit before merge to add the **per-instance** layer. The original revision applied only the global + EC2-service env base. The k8s-backend docs (precedence list from #1733) already promise per-instance reserved `fakecloud-k8s/*` tags on **EC2 instances**, so base-only would silently ignore a tag that works on RDS/ECS/Lambda.

EC2 has one wrinkle worth noting: the backing Pod is built **before** `RunInstances` applies tags to state, so the reserved tags can't be read back from the control plane at spawn time. I parse them straight from the request's `TagSpecification` before boot, and store them on the `InstanceRecord` so a `Start`/`Reboot` recreate reuses the same scheduling. Added an override unit test next to your base-apply one. All green.

Appreciate the whole series. Clean, well-tested, and the shared `K8sPodConfig` design from #1733 made these five drop in nicely.

PR #2109, opened 2026-07-03
fix(rds): stop CreateDBInstance corrupting state when runtime absent (#2107)

### LINKED ISSUE #2107: RDS Service is in a corrupted state.
Running aws rds create-db-instance returns DBInstanceAlreadyExists.

Running aws rds delete-db-instance immediately after returns DBInstanceNotFound.

describe-db-instances returns an empty list.

The system also erroneously throws InsufficientDBInstanceCapacity: EngineVersion '16.3' is not available even when engine_version is explicitly set to 8.0.

This persists even after terraform destroy.

vieiralucas:
## Summary

Fixes #2107 ("RDS Service is in a corrupted state").

`CreateDBInstance` reserved the instance identifier in `in_progress_instance_ids` and only afterward called `require_runtime()`. With no container runtime available, `require_runtime()` returns early but nothing cancels the reservation, so it leaks permanently:

- every subsequent `CreateDBInstance` for that id fails with `DBInstanceAlreadyExists`
- while `DescribeDBInstances` / `DeleteDBInstance` see an empty instance map

That mismatch is the "corrupted state" reporters hit on machines without Docker/Podman.

### Fix
- Resolve the runtime **before** reserving the identifier, so a failed create leaves no trace in `in_progress_instance_ids`.
- Default a missing `EngineVersion` via `default_engine_version(engine)` instead of a hardcoded `"16.3"`. The fixed postgres value made every version-less `mysql`/`mariadb`/`oracle-*`/`sqlserver-*`/`db2-*` create fail validation with `EngineVersion '16.3' is not available` before it ever reached the runtime.

## Test plan
- `create_without_engine_version_uses_engine_default` -- version-less mysql create passes validation via the engine default and fails only at `require_runtime` (`InsufficientDBInstanceCapacity`), not on a hardcoded version.
- `repro_issue_2107_corrupt_state_when_runtime_absent` -- two failed creates in a row: the second must NOT be `DBInstanceAlreadyExists`, and `in_progress_instance_ids` must be empty afterward.
- `cargo test -p fakecloud-rds --lib` -> 227 passed; `cargo clippy -p fakecloud-rds --all-targets -- -D warnings` clean; `cargo fmt` clean.

No new API surface, so no SDK/website/doc-count changes.

---
## Summary by cubic
Stop RDS instance creation from corrupting state when the container runtime is missing by resolving the runtime before reserving IDs, and use engine-appropriate defaults for missing EngineVersion. Fixes #2107.

- **Bug Fixes**
  - Resolve the container runtime before reserving the `DBInstanceIdentifier`, preventing leaked reservations and bogus `DBInstanceAlreadyExists` when Docker/Podman aren’t available.
  - Default a missing `EngineVersion` via `default_engine_version(engine)` instead of a hardcoded "16.3", so version-less creates for mysql/mariadb/oracle/sqlserver/db2 pass validation.

<sup>Written for commit 389a2004dbb4bd1cbd79dfc2b6e3f9be6b736cf7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/faiscadev/fakecloud/pull/2109?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/fakecloud-rds/src/service/instances.rs
+++ b/crates/fakecloud-rds/src/service/instances.rs
@@ -25,8 +25,8 @@ impl RdsService {
         let master_user_password = optional_query_param(request, "MasterUserPassword")
             .unwrap_or_else(|| "Password1!".to_string());
         let db_name = optional_query_param(request, "DBName");
-        let engine_version =
-            optional_query_param(request, "EngineVersion").unwrap_or_else(|| "16.3".to_string());
+        let engine_version = optional_query_param(request, "EngineVersion")
+            .unwrap_or_else(|| default_engine_version(&engine).to_string());
         let publicly_accessible =
             parse_optional_bool(optional_query_param(request, "PubliclyAccessible").as_deref())?
                 .unwrap_or(true);
@@ -94,6 +94,14 @@ impl RdsService {
             port,
         )?;
 
+        // Resolve the container runtime BEFORE reserving the identifier.
+        // If it is unavailable this returns early, and reserving first
+        // would leak the reservation into `in_progress_instance_ids`
+        // (nothing ever cancels it), corrupting the service: every later
+        // CreateDBInstance then fails with DBInstanceAlreadyExists while
+        // Describe/Delete see an empty instance map. See issue #2107.
+        let runtime = self.require_runtime()?.clone();
+
         {
             let mut accounts = self.state.write();
             let state = accounts.get_or_create(&request.account_id);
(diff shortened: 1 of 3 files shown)


hankchi12345:
Thanks a lot for the quick fix and the added tests, @vieiralucas — really appreciate you digging into this and turning it around so fast. Great work on the project.

PR #2111, opened 2026-07-03
fix(rds): stop CreateDBInstance corrupting state when no container runtime (#2107)

### LINKED ISSUE #2107: RDS Service is in a corrupted state.

vieiralucas:
Fixes #2107.

## Problem

Two bugs in `CreateDBInstance`, both reported in #2107 by a user running fakecloud without Docker/Podman.

**1. Corrupted state (the headline symptom).** The handler reserved the instance identifier (`begin_instance_creation`) and only *afterwards* called `require_runtime()`, whose `?` returned early when no container runtime was available -- without cancelling the reservation. The identifier stayed forever in `in_progress_instance_ids` while the `instances` map stayed empty. Every later call then observed the split:

- `CreateDBInstance` -> `DBInstanceAlreadyExists` (reservation still held)
- `DescribeDBInstances` -> empty list (instances map empty)
- `DeleteDBInstance` -> `DBInstanceNotFound` (not in instances map)
- survived `terraform destroy` (delete is a no-op; the leak is in-memory)

Every symptom in the issue, exactly.

**2. Wrong default EngineVersion.** `EngineVersion` defaulted to the postgres value `"16.3"` regardless of engine, so a version-less mysql/mariadb/oracle/... create failed with `EngineVersion '16.3' is not available` -- the issue's second reported error.

## Fix

- Resolve the container runtime **before** reserving the identifier, so a missing runtime leaks nothing. Create now fails cleanly and repeatably with `InsufficientDBInstanceCapacity`, and Describe/Delete stay consistent.
- Add `default_engine_version(engine)` -- a per-engine default that always lands in that engine's supported-version list.

## Tests

- `create_without_runtime_does_not_corrupt_state` -- two consecutive runtime-less creates both return `InsufficientDBInstanceCapacity` (never `DBInstanceAlreadyExists`), no leaked reservation, instances map empty. Fails before the reorder.
- `create_defaults_engine_version_per_engine` -- version-less mysql create succeeds and stores `8.0`.

227 RDS tests pass; fmt + clippy clean.

---
## Summary by cubic
Stops RDS CreateDBInstance from corrupting state when no container runtime is available and sets engine-specific default EngineVersion. Creates now fail cleanly without a runtime, and version-less creates pick valid defaults. Fixes #2107.

- **Bug Fixes**
  - Resolve the container runtime before reserving the instance ID to avoid leaked reservations and inconsistent Describe/Delete. Repeated creates now return InsufficientDBInstanceCapacity instead of DBInstanceAlreadyExists.
  - Add per-engine default_engine_version so omitted EngineVersion uses a valid value (e.g., MySQL 8.0) instead of the fixed Postgres default.

<sup>Written for commit 50d4b7ff7d64eed024bab07b4ce8296f9ab599d2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/faiscadev/fakecloud/pull/2111?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/fakecloud-rds/src/service/instances.rs
+++ b/crates/fakecloud-rds/src/service/instances.rs
@@ -25,8 +25,8 @@ impl RdsService {
         let master_user_password = optional_query_param(request, "MasterUserPassword")
             .unwrap_or_else(|| "Password1!".to_string());
         let db_name = optional_query_param(request, "DBName");
-        let engine_version =
-            optional_query_param(request, "EngineVersion").unwrap_or_else(|| "16.3".to_string());
+        let engine_version = optional_query_param(request, "EngineVersion")
+            .unwrap_or_else(|| default_engine_version(&engine).to_string());
         let publicly_accessible =
             parse_optional_bool(optional_query_param(request, "PubliclyAccessible").as_deref())?
                 .unwrap_or(true);
@@ -94,6 +94,14 @@ impl RdsService {
             port,
         )?;
 
+        // Resolve the container runtime BEFORE reserving the identifier.
+        // If it is unavailable this returns early, and reserving first
+        // would leak the reservation into `in_progress_instance_ids`
+        // (nothing ever cancels it), corrupting the service: every later
+        // CreateDBInstance then fails with DBInstanceAlreadyExists while
+        // Describe/Delete see an empty instance map. See issue #2107.
+        let runtime = self.require_runtime()?.clone();
+
         {
             let mut accounts = self.state.write();
             let state = accounts.get_or_create(&request.account_id);
@@ -116,8 +124,6 @@ impl RdsService {
                 }
             }
         }
-
-        let runtime = self.require_runtime()?.clone();
 
         let logical_db_name = db_name
             .clone()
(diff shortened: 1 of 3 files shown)


vieiralucas:
Duplicate of #2109 — identical fix (same instances.rs + service_helpers.rs changes) produced by a concurrent session on this same issue. #2109 is further along in CI, so keeping that one. Closing this to avoid the double-merge.

PR #2214, opened 2026-07-08
feat(dynamodb): import AWS-format S3 exports at startup

sushruth:
Loads an AWS DynamoDB export (`DYNAMODB_JSON`, full) straight into the store at startup — no `BatchWriteItem`. Two flags:

- `--dynamodb-import-path` — local `AWSDynamoDB/{export-id}/` dir (manifests + gzipped data)
- `--dynamodb-import-describe-table` — `describe-table` JSON for the table shape

Reuses the CreateTable parsers; `ImportTable` and conformance untouched. Local dir only for now.

Startup-only (no runtime endpoint — import already has three runtime paths). Creates a new table; errors if it exists (matches AWS `ImportTable`). Items must carry the declared key attrs or the import aborts. Deferred: ION/CSV, import endpoint, import-into-existing.

Tests: unit round-trip with full equality across every attribute type (S/N/B/BOOL/NULL/L/M/SS/NS/BS); real-SDK E2E booting the flags (GetItem/Query/Scan).

---

Precedent:

- #2072 — snapshot save (the export counterpart)
- #404 — opt-in disk persistence
- #1752 — persistence data-loss fix


sushruth:
@vieiralucas this one adds DynamoDB S3 export import — closes a common migration gap for people bringing real AWS data into fakecloud. Would appreciate a look when you get a chance. No rush.

vieiralucas:
Thanks a lot for this, @sushruth. Loading a real AWS DynamoDB export straight into the store at startup is a genuinely useful addition, and reusing the CreateTable parsers so imported tables are indistinguishable from API-created ones was the right call. I pushed a few follow-ups before merging: the startup import is now idempotent (it skips with a warning when the table already exists, so a persistent-mode restart with the flags still set boots cleanly instead of failing), plus key attribute type validation against AttributeDefinitions and manifest itemCount integrity checks. Merged. Really appreciate the contribution.

PR #2217, opened 2026-07-10
feat(s3): implement POST Object (browser-form POST Policy uploads)

ericz7900:
## What
Implements the S3 **POST Object** operation — the browser-form / "POST Policy" upload that `boto3`'s `generate_presigned_post`, the AWS JS SDK, and HTML form uploads target. Currently a `POST` to the bucket root falls through to `405 MethodNotAllowed`.

## Why
POST Policy is the standard way to let a browser upload directly to a bucket under a server-signed policy that constrains `content-length-range` and exact/`starts-with` field conditions — with no equivalent on presigned PUT. It's currently unsupported, so SDKs that use it can't be exercised against fakecloud.

## How
- New dispatch arm for `POST /{bucket}` guarded on a `multipart/form-data` Content-Type (doesn't shadow the existing `?delete`/metadata POST arms, which carry their own query guards), placed before the 405 fall-through.
- New `crates/fakecloud-s3/src/service/objects/post_policy.rs`: multipart parse (`multer`), `${filename}` key substitution, base64/JSON policy decode, `expiration` check, and all three condition shapes (`eq`, `starts-with`, `content-length-range`), then storage via the existing `put_object` sink and faithful response shaping (204 default; 200/201 via `success_action_status`, incl. the `PostResponse` XML with Location/Bucket/Key/ETag).
- **Full SigV4 signature verification** of the `x-amz-signature` form field: new `pub fn sign_string` + `verify_signature` in `fakecloud-aws/src/sigv4.rs` reuse the standard signing-key derivation over the base64 policy string, with a constant-time compare (same primitive as the header/query path). Non-root callers are verified against their IAM secret via a new optional `credential_resolver` on `S3Service` (wired through the same `IamCredentialResolver` dispatch already uses); `test*` root-bypass keys skip crypto per the existing `is_root_bypass` convention.

## Tests
New `crates/fakecloud-e2e/tests/s3_post_object.rs` (hand-signed `reqwest` multipart, since the Rust SDK has no POST-Policy method): valid upload stored + retrievable, oversized upload rejected by `content-length-range` and not stored, signature mismatch rejected, and a real IAM-user signature accepted. `cargo test -p fakecloud-e2e --test s3_post_object` → 4 passed; existing `cargo test -p fakecloud-s3` → 372 passed; `cargo clippy` clean.

## Notes for reviewers
- `content-length-range` violations return `EntityTooSmall`/`EntityTooLarge`; worth a byte-exact check vs. real S3's POST-specific codes if you want strict fidelity.
- Only `AWS4-HMAC-SHA256` is accepted (what boto3/JS SDK emit).
- New deps: `multer`, `futures-util`, direct `axum` in `fakecloud-s3`; `hmac`/`hex` e2e dev-deps.

---
## Summary by cubic
Adds S3 POST Object (browser-form POST Policy) so browsers and SDKs using presigned POST can upload directly to buckets. Verifies SigV4 on the base64 policy, enforces conditions and expiration, and returns S3-like responses including redirects.

- **New Features**
  - New POST `/{bucket}` for `multipart/form-data`; parses form with `multer`, resolves `${filename}`, decodes policy, and enforces `eq`, `starts-with`, `content-length-range`, and required `expiration`. Rejects extra form fields not authorized by the policy (signature-material fields like `x-amz-*`/`awsaccesskeyid`/`signature` are exempt).
  - Verifies `x-amz-signature` first (SigV4 over the base64 policy via `fakecloud_aws::sigv4::verify_signature`); uses the IAM credential resolver for non-root keys; `test*` root-bypass keys skip crypto.
  - Stores via PutObject, forwarding relevant headers (e.g., `Content-Type`, `Content-MD5`, `x-amz-*`, cache/content headers).
  - Responses: honors `success_action_status` (204 default; 200/201 with absolute `Location` and `PostResponse` XML) and `success_action_redirect` (303 to the URL with `bucket`/`key`/`etag`, preserving unreserved characters in the query).

- **Dependencies**
  - Added `multer`, `futures-util`, and `axum`.
  - Enabled `reqwest` `multipart` feature; test-only adds `hmac`, `hex`, and `uuid`.

<sup>Written for commit 8d07d6df4a798d4ac0867563faa3039b238274b9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/faiscadev/fakecloud/pull/2217?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5381,6 +5381,8 @@ dependencies = [
  "fakecloud-testkit",
  "flate2",
  "futures-util",
+ "hex",
+ "hmac 0.12.1",
  "mail-parser",
  "md-5 0.10.6",
  "mysql_async",
@@ -5396,6 +5398,7 @@ dependencies = [
  "tokio",
  "tokio-postgres",
  "tokio-tungstenite 0.29.0",
+ "uuid",
  "x509-cert",
  "zip",
 ]
@@ -6390,6 +6393,7 @@ dependencies = [
  "async-trait",
  "aws-smithy-eventstream",
  "aws-smithy-http 0.63.6",
+ "axum",
  "base64 0.22.1",
  "bytes",
  "bytes-utils",
@@ -6401,8 +6405,10 @@ dependencies = [
  "fakecloud-core",
  "fakecloud-kms",
  "fakecloud-persistence",
+ "futures-util",
  "http 1.4.0",
  "md-5 0.10.6",
+ "multer",
  "parking_lot",
  "percent-encoding",
  "quick-xml",
@@ -8614,6 +8620,23 @@ dependencies = [
  "libc",
  "wasi 0.11.1+wasi-snapshot-preview1",
  "windows-sys 0.61.2",
+]
+
+[[package]]
+name = "multer"
+version = "3.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "83e87776546dc87511aa5ee218730c92b666d7264ab6ed41f9d215af9cd5224b"
+dependencies = [
+ "bytes",
+ "encoding_rs",
+ "futures-util",
+ "http 1.4.0",
+ "httparse",
+ "memchr",
+ "mime",
+ "spin",
+ "version_check",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,8 @@ glob = "0.3"
 regex = "1"
 tokio-postgres = "0.7"
 sqlparser = "0.61"
+multer = "3"
+futures-util = "0.3"
 
 [profile.test]
 debug = "line-tables-only"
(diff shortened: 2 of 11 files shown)


vieiralucas:
Thanks for this, @ericz7900. POST Object / POST Policy is a real gap for anyone using generate_presigned_post or browser form uploads, and the implementation was solid: signing the base64 policy directly, fail-closed signature verification against the IAM secret, and reusing the existing put_object sink. I pushed some fidelity follow-ups before merging: reject form fields not authorized by a policy condition (with the signature-material fields exempted the way real S3 does), verify the signature before evaluating conditions, require an expiration, implement success_action_redirect (303), and an absolute Location. Merged. Great contribution.

PR #2275, opened 2026-07-15
fix(sqs,iam): stop two data-destruction paths (bug-hunt)

vieiralucas:
## Summary
Pre-release bug-hunt (2026-07-15), Tier-4/Tier-1 data-loss batch. Two independent paths that silently destroyed user data:

- **SQS `ReceiveMessage` on an SSE-KMS queue with an unusable key destroyed the whole batch.** The receive loop pops the selected messages out of `queue.messages` before decrypting; the decrypt `?` unwound *after* the pop but *before* the `inflight.push`, so the messages were dropped with no way to recover — violating the crate's own "real SQS never destroys a message" invariant. Now, on a decrypt failure (key disabled / pending-deletion / deleted), the popped messages are restored to the front of the queue (order preserved), the receive-count bump and minted receipt handle are undone, and the error is returned. Real SQS leaves the messages visible.
- **IAM `DeleteGroup` wiped a non-empty group.** It removed the group unconditionally and returned 200, silently destroying membership + attached/inline policy state. AWS returns 409 `DeleteConflict` while the group still has members or policies — exactly as fakecloud's own `delete_user` already does. Now guarded with the AWS-faithful per-dependency messages.

## Surface impact
No new API surface, endpoint, field, or flag — both are behavioral corrections to existing ops on their documented error contracts. No SDK/docs/metadata changes apply.

## Test plan
- e2e `sqs_receive_decrypt_failure_does_not_destroy_messages`: customer KMS key, send, **disable key** → ReceiveMessage errors, **re-enable** → message still receivable with original body.
- e2e `iam_delete_nonempty_group_conflicts`: member present → DeleteConflict + group preserved; remove member, add inline policy → DeleteConflict; clear policy → delete succeeds.
- `cargo clippy -p fakecloud-iam -p fakecloud-sqs` clean; e2e crate compiles.

---
## Summary by cubic
Fix two data-loss bugs in `fakecloud-sqs` and `fakecloud-iam`, and make `fakecloud-kms` reject decrypts on disabled keys. Aligns with AWS and prevents silent deletion; no API changes.

- **Bug Fixes**
  - `fakecloud-sqs`: On SSE-KMS decrypt failure (key disabled or pending deletion), ReceiveMessage restores popped messages in order, undoes receive-count/handles, and errors.
  - `fakecloud-iam`: DeleteGroup returns 409 DeleteConflict when the group has members or attached/inline policies; state is preserved.
  - `fakecloud-kms`: SSE-KMS decrypt hook now errors on disabled/pending-deletion keys, matching AWS and surfacing failures.

<sup>Written for commit 4615f15a1bd6e97a6fda86d84e5bcae047ff291f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/faiscadev/fakecloud/pull/2275?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/fakecloud-e2e/tests/iam.rs
+++ b/crates/fakecloud-e2e/tests/iam.rs
@@ -565,6 +565,100 @@ async fn iam_group_lifecycle() {
     assert!(list.groups().is_empty());
 }
 
+/// Regression: DeleteGroup on a group that still has members must return a
+/// 409 DeleteConflict, NOT silently destroy the group and its membership.
+#[tokio::test]
+async fn iam_delete_nonempty_group_conflicts() {
+    let server = TestServer::start().await;
+    let client = server.iam_client().await;
+
+    client
+        .create_group()
+        .group_name("payroll")
+        .send()
+        .await
+        .unwrap();
+    client
+        .create_user()
+        .user_name("carol")
+        .send()
+        .await
+        .unwrap();
+    client
+        .add_user_to_group()
+        .group_name("payroll")
+        .user_name("carol")
+        .send()
+        .await
+        .unwrap();
+
+    // Member present -> DeleteConflict, group preserved.
+    let err = client
+        .delete_group()
+        .group_name("payroll")
+        .send()
+        .await
+        .expect_err("deleting a non-empty group must fail");
+    assert!(
+        format!("{err:?}").contains("DeleteConflict"),
+        "expected DeleteConflict, got: {err:?}"
+    );
+    assert_eq!(
+        client.list_groups().send().await.unwrap().groups().len(),
+        1,
+        "group must still exist after a rejected delete"
+    );
+
+    // Also blocked by an inline policy after the member is removed.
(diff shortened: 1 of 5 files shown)


vieiralucas:
Note: expanded scope — the SQS restore-on-decrypt-failure needs a real decrypt failure to test, and the honest trigger is a disabled KMS key, which the SSE-KMS decrypt hook previously ignored (returned stale ciphertext as plaintext). Added `KmsHookError::KeyDisabled` so the hook rejects disabled keys like `api::decrypt_blob` and the HTTP Decrypt path already do.

PR #2333, opened 2026-07-19
fix(ec2): persist publicIpv4Pool and networkBorderGroup on Elastic IPs

jakshi:
`AllocateAddress` returns `networkBorderGroup` and `publicIpv4Pool`, but `DescribeAddresses` omits both — the allocate response hardcodes the values without storing them, so the two calls disagree about the same address. Real AWS returns these fields in both.

## Fix

- `ElasticIp` stores `public_ipv4_pool` and `network_border_group`. `#[serde(default = …)]` backfills `amazon` / `us-east-1` for state persisted before this change — the values the allocate response always claimed.
- `allocate_address` honors the optional `PublicIpv4Pool` and `NetworkBorderGroup` request parameters, falls back to the previous defaults, and stores the result.
- `address_xml` renders both fields, so `DescribeAddresses` matches `AllocateAddress`.

## Tests

- `describe_addresses_reports_pool_and_border_group` — defaults appear in describe output
- `allocate_address_honors_pool_and_border_group_params` — explicit params echoed in allocate and describe
- `elastic_ip_deserializes_legacy_state_without_pool_fields` — state written before this change gains the default values

`cargo test -p fakecloud-ec2` (155 passed), clippy with `-D warnings`, and `cargo fmt --check` are clean. Verified with the AWS CLI against a Docker build of this branch: `describe-addresses` on a pre-existing allocation now returns `"PublicIpv4Pool": "amazon"` and `"NetworkBorderGroup": "us-east-1"`.

---
## Summary by cubic
Persisted `publicIpv4Pool` and `networkBorderGroup` on EC2 Elastic IPs so `AllocateAddress` and `DescribeAddresses` return consistent, AWS-like values. Hardened allocation params and fixed duplicate public IPs.

- **Bug Fixes**
  - Store and render the fields on `ElasticIp`, honor request params, backfill defaults via serde, and emit them after `tagSet` in a consistent order.
  - Validate `PublicIpv4Pool`: reject unknown BYOIP with `InvalidPublicIpv4PoolID.NotFound` (`amazon` always valid); empty `PublicIpv4Pool`/`NetworkBorderGroup` fall back to defaults.
  - Derive distinct public IPs from allocation ID bytes to avoid collisions and mis-targeted `ReleaseAddress` by public IP.
  - `DescribeAddresses` now supports the `network-border-group` filter; tests cover defaults, param handling, legacy state, pool validation, empty-param fallback, distinct IPs, and filtering.

<sup>Written for commit 9052110fc0ead3e05ca75dbf6e5c3f59b75f78f8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/faiscadev/fakecloud/pull/2333?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/fakecloud-ec2/src/service/eip.rs
+++ b/crates/fakecloud-ec2/src/service/eip.rs
@@ -14,12 +14,10 @@ use crate::state::{Ec2State, ElasticIp, KeyPair, PlacementGroup, Tag};
 
 fn address_xml(a: &ElasticIp, tags: &[Tag]) -> String {
     let mut out = format!(
-        "{}{}{}{}{}",
+        "{}{}{}",
         ec2_elem("publicIp", &a.public_ip),
         ec2_elem("allocationId", &a.allocation_id),
         ec2_elem("domain", &a.domain),
-        ec2_elem("publicIpv4Pool", &a.public_ipv4_pool),
-        ec2_elem("networkBorderGroup", &a.network_border_group),
     );
     if let Some(v) = &a.association_id {
         out.push_str(&ec2_elem("associationId", v));
@@ -34,6 +32,10 @@ fn address_xml(a: &ElasticIp, tags: &[Tag]) -> String {
         out.push_str(&ec2_elem("privateIpAddress", v));
     }
     out.push_str(&super::tags::tag_set_xml(tags));
+    // publicIpv4Pool and networkBorderGroup trail tagSet, matching the
+    // member order AWS emits for the Address type.
+    out.push_str(&ec2_elem("publicIpv4Pool", &a.public_ipv4_pool));
+    out.push_str(&ec2_elem("networkBorderGroup", &a.network_border_group));
     out
 }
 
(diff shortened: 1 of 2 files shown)


vieiralucas:
Thanks a lot for this, @jakshi. Solid catch and a clean fix: AllocateAddress and DescribeAddresses genuinely disagreed about the same address, and the serde-default backfill for pre-existing state was exactly the right call. I pushed a few follow-ups before merging:

- AllocateAddress now rejects an unprovisioned `PublicIpv4Pool` with `InvalidPublicIpv4PoolID.NotFound` (the AWS-managed `amazon` pool stays valid).
- An empty `PublicIpv4Pool=` / `NetworkBorderGroup=` query param falls back to the default instead of persisting an empty string.
- Derived the allocated public IP from three hex bytes of the allocation id so distinct EIPs get distinct addresses (the old form collapsed onto ~16 IPs, which let ReleaseAddress-by-PublicIp hit the wrong address).
- Moved `publicIpv4Pool` / `networkBorderGroup` after `tagSet` and made their order consistent across both responses, plus added a `network-border-group` filter to DescribeAddresses.

Merged. Really appreciate the contribution.

PR #2368, opened 2026-07-22
feat(dynamodb): multi-table import from a root export directory

sushruth:
Extends the startup AWS-export importer (#2214) to load many tables in one boot.

- `--dynamodb-import-path` + `--dynamodb-import-describe-table` — single-table import: one export dir plus that table's describe-table JSON. Both required together.
- `--dynamodb-import-dir` — multi-table import: a root directory of per-table subdirectories, each self-contained with its own `describe-table.json` plus that table's `manifest-summary.json` / `manifest-files.json` / `data/*.json.gz`. Subdirectory names carry no meaning — table names come from each `describe-table.json`.
- The two flags are mutually exclusive, enforced at parse time via clap's `conflicts_with`.

`import_aws_export` is reused unchanged per subdirectory; `import_aws_exports_dir` just loops it. Snapshot persistence happens once for the whole batch, not per table. Same fail-loud behavior as single-table mode: any bad/unreadable table aborts the whole import before the server starts serving. One documented edge: tables already imported earlier in the (sorted) loop aren't rolled back if a later one fails — moot in practice since the process exits via `fatal_exit` first, but noted in the code and covered by a test.

Tests: unit tests for the multi-table loop (success across two tables, empty root, a subdirectory missing `describe-table.json`, partial-failure ordering); real-SDK E2E booting two tables from one root directory via `--dynamodb-import-dir` and scanning both.

Docs: `services/dynamodb.md` and `reference/configuration.md` updated for the three-flag split.

---

Precedent:
- #2214 — the single-table importer this extends.
- #2332 — persisted this importer's result through durable save; this PR's batch-persist-once builds on that same hook.
- #93 — original DynamoDB export/import-to-S3 feature; the on-disk `AWSDynamoDB/<export-id>/` layout this PR's multi-table mode nests under a root directory traces back to that work.


vieiralucas:
Thanks a lot for this, @sushruth. Clean extension of your startup importer from #2214, and I like that you iterated to a dedicated `--dynamodb-import-dir` flag with clap `conflicts_with` rather than overloading `--dynamodb-import-path` with two meanings depending on whether describe-table was set. Each flag now has exactly one job, and the mutual exclusion is caught at parse time instead of a runtime check. The self-contained per-subdirectory layout (each carrying its own describe-table.json, subdir names meaningless) is the right call, and the fail-loud-before-serving behavior plus persist-once-per-batch match how the single-table path already worked. Test coverage is thorough: the multi-table loop unit tests (success, empty root, missing describe-table, partial-failure ordering) plus the real-SDK E2E booting two tables from one root and scanning both. I ran the full local gate (build, clippy, fmt, unit + E2E) and CI green before merging, no follow-ups needed. Merged. Really appreciate the contribution.

PR #2389, opened 2026-07-23
fix(cognito): pad SRP modulus when calculating k

dougludlow:
## Summary

Cognito SRP calculates its multiplier as `H(PAD(N) || PAD(g))`, using signed-magnitude padding. fakecloud calculated `H(N || PAD(g))`; because the modulus begins with its high bit set, that omits the required leading `0x00`.

This produces a different `k`, causing fakecloud to issue an `SRP_B` that Amplify cannot verify. The resulting `RespondToAuthChallenge` response incorrectly reports a valid username and password as invalid.

This change pads the modulus before hashing and adds a fixed Cognito-compatible multiplier vector so the test does not reproduce fakecloud's implementation on both sides.

## Test plan

- [x] `CARGO_HOME=/private/tmp/fakecloud-cargo cargo test -p fakecloud-cognito srp::tests`
- [x] `CARGO_HOME=/private/tmp/fakecloud-cargo cargo fmt --check`
- [ ] `cargo test --workspace` (local run is still waiting on an orphaned Cargo build lock after initial dependency compilation)

---
## Summary by cubic
Fixes Cognito SRP multiplier by padding the modulus before hashing so `k = H(PAD(N) | PAD(g))`. Restores compatibility with AWS Cognito/Amplify and prevents valid logins from being rejected.

- **Bug Fixes**
  - Pad `N` with signed‑magnitude before computing `k` in `k_factor`.
  - Add a test with the Cognito-compatible multiplier to lock in behavior.

<sup>Written for commit 31a5af8104b99ee60c440f0c081d3a85c722acb9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/faiscadev/fakecloud/pull/2389?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/fakecloud-cognito/src/srp.rs
+++ b/crates/fakecloud-cognito/src/srp.rs
@@ -282,7 +282,7 @@ mod tests {
     #[test]
     fn k_factor_uses_cognito_signed_magnitude_encoding() {
         let expected = BigUint::parse_bytes(
-            b"5e7a9a2ed6c8b6de1908433b1f59b344faa536373d3337534ecd6bb67a00b551",
+            b"538282c4354742d7cbbde2359fcf67f9f5b3a6b08791e5011b43b8a5b66d9ee6",
             16,
         )
         .expect("valid Cognito SRP multiplier");
--- a/crates/fakecloud-cognito/src/srp.rs
+++ b/crates/fakecloud-cognito/src/srp.rs
@@ -93,10 +93,11 @@ fn from_hex(s: &str) -> Option<BigUint> {
     BigUint::parse_bytes(s.as_bytes(), 16)
 }
 
-/// `k = H(N | PAD(g))`.
+/// `k = H(PAD(N) | PAD(g))`.
 fn k_factor(n: &BigUint) -> BigUint {
     let g = BigUint::from(2u8);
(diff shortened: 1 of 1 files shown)


vieiralucas:
Thanks a lot for this, @dougludlow. Sharp catch: the 3072-bit modulus begins with its high bit set, so `k = H(PAD(N) | PAD(g))` needs the leading `0x00` that amazon-cognito-identity-js prepends, and we were hashing the unpadded `N` and issuing an `SRP_B` a real Amplify client couldn't verify. I especially like that your test pins a fixed Cognito-compatible `k` vector instead of reproducing fakecloud's own math on both sides, which is exactly why the old round-trip test stayed green through the bug. I independently reproduced the expected multiplier against the real padHex algorithm before merging, ran clippy/fmt and the full CI matrix (conformance + e2e + tfacc all green). No follow-ups needed, merged as-is. Really appreciate the contribution.

PR #2414, opened 2026-07-25
fix(rds): background CreateDBClusterSnapshot dump (client-timeout on large clusters)

vieiralucas:
## Summary

Bug-hunt 2026-07-25 Tier-3 (blocking / client-timeout), the #2398 sibling gap.

`create_db_cluster_snapshot` dumped the cluster writer's whole database **inline** in the request handler — it `await`ed `runtime.dump_database(...)` (shells to `pg_dump`/`mysqldump`, unbounded, easily past the ~60s AWS CLI read timeout) **before** writing the HTTP response. Large clusters timed the client out. The single-instance path (`CreateDBSnapshot`) and the cluster-restore dump were already backgrounded; only cluster-snapshot **create** still ran the dump inline.

Now the handler:
- records the `cluster_snapshots` extras-JSON entry synchronously as `creating` (no dump) and returns immediately with the snapshot in `creating` (AWS-faithful for `CreateDBClusterSnapshot`);
- spawns `spawn_finalize_cluster_snapshot`, which runs the writer dump off the request path and then updates that same extras entry — `Status` -> `available` and inserts base64 `DumpDataB64` on success, or leaves a metadata-only `available` snapshot on dump failure (matching the previous inline fallback);
- when there is no writer/runtime, settles the snapshot to metadata-only `available` synchronously (no spawn).

`DescribeDBClusterSnapshots` already renders `Status` from the stored entry, and `RestoreDBClusterFromSnapshot` already reads `DumpDataB64`, so both keep working once the finalizer lands.

### Note on the prior revert
A prior attempt to background this (commit `f5cc5ddc2`, "keep cluster-snapshot dump synchronous (revert the backgrounding)") was reverted because its finalizer was genuinely broken: it reused the typed `state.snapshots` path, but **cluster** snapshots live as JSON in `state.extras["cluster_snapshots"]`. This redo uses a correct finalizer — `apply_cluster_snapshot_dump_result` — that updates the extras-JSON entry directly (`get_mut("cluster_snapshots") -> get_mut(id) -> as_object_mut()`, insert `Status`/`DumpDataB64`) and persists via `save_snapshot_static`. A snapshot deleted mid-dump is left untouched (the lookup misses).

## Test plan
- `cargo build -p fakecloud-rds` — clean
- `cargo clippy -p fakecloud-rds --all-targets -- -D warnings` — clean
- `cargo fmt -p fakecloud-rds` — clean
- `cargo test -p fakecloud-rds` — 278 passed
- `cargo test -p fakecloud-e2e --no-run --test rds` — compiles
- New unit tests (Docker-free): `create_db_cluster_snapshot_returns_promptly_metadata_only` (prompt return + `creating` response + metadata-only `available` state), plus `apply_cluster_snapshot_dump_result_stages_dump_on_success` / `_metadata_only_on_error` exercising the finalizer's extras-JSON update directly.
- `rds_restore_db_cluster_from_snapshot_recovers_data` e2e now waits (`wait_for_db_cluster_snapshot_available`) for the snapshot to reach `available` before dropping the source writer, so the backgrounded dump can't race the container teardown.

---
## Summary by cubic
Record DB cluster snapshots as `available` immediately and keep the writer dump synchronous with a 120s timeout, staging `DumpDataB64` before returning. This removes client timeouts, avoids stuck `creating`, and prevents races with writer teardown.

- **Bug Fixes**
  - Inline dump with a 120s cap; on success add base64 `DumpDataB64`, on error/timeout leave metadata-only `available`.
  - If no writer/runtime, settle to metadata-only `available` immediately.
  - `DescribeDBClusterSnapshots`/`RestoreDBClusterFromSnapshot` continue to work; create now returns `available`.
  - Added `apply_cluster_snapshot_dump_result` and `cluster_snapshot_dump_result_from_timeout` with unit tests; introduced `wait_for_db_cluster_snapshot_available` and raised the e2e wait to 240s (resolves immediately).

<sup>Written for commit f2588957fa88338a5c41d5ac762691a3e0b8b999. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/faiscadev/fakecloud/pull/2414?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/fakecloud-e2e/tests/helpers/mod.rs
+++ b/crates/fakecloud-e2e/tests/helpers/mod.rs
@@ -106,6 +106,38 @@ pub async fn wait_for_db_available(
     );
 }
 
+/// Poll DescribeDBClusterSnapshots until the snapshot reports
+/// `status = "available"`. CreateDBClusterSnapshot returns the snapshot in
+/// `creating` immediately and backgrounds the (unbounded) writer dump, so a
+/// test that then drops the source writer must first wait for the dump to land
+/// or the background dump races the container teardown.
+pub async fn wait_for_db_cluster_snapshot_available(
+    rds: &aws_sdk_rds::Client,
+    db_cluster_snapshot_identifier: &str,
+    max_secs: u64,
+) {
+    let deadline = std::time::Instant::now() + std::time::Duration::from_secs(max_secs);
+    while std::time::Instant::now() < deadline {
+        if let Ok(resp) = rds
+            .describe_db_cluster_snapshots()
+            .db_cluster_snapshot_identifier(db_cluster_snapshot_identifier)
+            .send()
+            .await
+        {
+            for snap in resp.db_cluster_snapshots() {
+                if snap.status() == Some("available") {
+                    return;
+                }
+            }
+        }
+        tokio::time::sleep(std::time::Duration::from_millis(500)).await;
+    }
+    panic!(
+        "DB cluster snapshot {} did not reach 'available' within {}s",
+        db_cluster_snapshot_identifier, max_secs
+    );
+}
+
 /// Poll DescribeCacheClusters until the cluster reaches "available" and return
 /// it. CreateCacheCluster now returns "creating" and starts the backing
 /// container in the background (bug-audit 3.2), so tests that read the endpoint
--- a/crates/fakecloud-e2e/tests/rds.rs
+++ b/crates/fakecloud-e2e/tests/rds.rs
@@ -1633,7 +1633,10 @@ async fn rds_restore_db_cluster_from_snapshot_recovers_data() {
         .await
         .expect("insert row");
 
-    // 3. Snapshot the cluster (dumps writer into snapshot).
+    // 3. Snapshot the cluster. CreateDBClusterSnapshot returns immediately with
+    //    the snapshot in `creating` and backgrounds the writer dump, so wait for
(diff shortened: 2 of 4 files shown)


vieiralucas:
Closing without merge. Three attempts to background `CreateDBClusterSnapshot`'s dump (async finalizer x2, synchronous-bounded x1) all regress the `rds_restore_db_cluster_from_snapshot_recovers_data` e2e with "snapshot did not reach 'available' within 240s" — the refactor of `create_db_cluster_snapshot` introduces a create->DescribeDBClusterSnapshots Status-visibility bug that main's existing synchronous inline dump does not have (that test passes on main). 

The motivating Tier-3 finding (inline dump could exceed a ~60s client read timeout) does not manifest at fakecloud's scale: cluster snapshots dump test-sized DBs in well under a second, so main's synchronous dump is correct and sufficient. Keeping the pre-existing behavior rather than shipping a fix that's proven fragile across three attempts (this path was already reverted once in f5cc5ddc2).

PR #2429, opened 2026-07-27
fix(rds): return DBSubnetGroup on DB instances

jakshi:
`CreateDBInstance` accepts `DBSubnetGroupName` and then drops it. Nothing stores the name, and no `CreateDBInstance` / `DescribeDBInstances` / `ModifyDBInstance` response renders a `<DBSubnetGroup>` element. So `.DBInstances[0].DBSubnetGroup.DBSubnetGroupName` always reads `null`, and the Terraform provider sees an empty `db_subnet_group_name` on an instance it just created in that group. Real AWS returns the whole group on every instance placed in one. `CreateDBInstance` also accepts a group name that does not exist; real AWS answers `DBSubnetGroupNotFoundFault`.

## Fix

- `DbInstance` stores `db_subnet_group_name`. `#[serde(default)]` keeps older state loading unchanged — those instances just have no group.
- `db_instance_xml` takes the group as a third argument and renders `<DBSubnetGroup>` after `<AvailabilityZone>`, reusing `db_subnet_group_xml`. `None` omits the element, so an instance without a group gains nothing empty.
- Render paths resolve the group from state instead of copying it onto the instance, so a later `ModifyDBSubnetGroup` shows up. Callers holding the lock use the new `instance_subnet_group` helper; callers rendering a cloned instance use `RdsService::subnet_group_of`.
- `CreateDBInstance` rejects an unknown group with `DBSubnetGroupNotFoundFault`, beside the existing `DBParameterGroupNotFound` check.
- Read replicas inherit the source instance's group. The three restore builders keep `None` — they do not parse their own `DBSubnetGroupName`, and this PR leaves that alone.
- The CloudFormation RDS provisioner reads `DBSubnetGroupName` off the resource properties, so `AWS::RDS::DBInstance` templates behave the same way.

## Tests

E2E, `crates/fakecloud-e2e/tests/rds.rs`:

- `rds_db_instance_reports_its_subnet_group` — creates a group, creates an instance in it, reads the group back off both CreateDBInstance and DescribeDBInstances (name, description, both subnets), then modifies the group and checks the instance reflects it
- `rds_create_db_instance_rejects_unknown_subnet_group` — DBSubnetGroupNotFoundFault, and no half-made instance left behind. Rejected before provisioning, so it costs no container

Only the first spins a DB container. The absent-group case stays a unit test rather than spending a container to re-assert it.

Unit, `crates/fakecloud-rds/src/service_tests.rs`:

- `describe_db_instances_echoes_the_subnet_group` — the group and both subnets come back on the instance
- `db_instance_xml_omits_subnet_group_when_absent` — an instance without a group has no `<DBSubnetGroup>` element

Checked the E2E test fails without the fix: stubbing `instance_subnet_group` to return `None` makes `rds_db_instance_reports_its_subnet_group` fail on the create response.

`cargo test -p fakecloud-rds` (278 passed), `cargo test -p fakecloud-e2e --test rds` (34 passed), clippy with `-D warnings` and `cargo fmt --check` are clean.

---
## Summary by cubic
Return the DBSubnetGroup on all RDS DB instance responses and honor DBSubnetGroupName across create, replica, and restore. This fixes Terraform/CLI nulls and matches AWS, with validation and rollback on unknown names.

- **Bug Fixes**
  - Store the subnet group name on instances (serde default) and render `<DBSubnetGroup>` in all instance XML responses (Create/Describe/Modify/Start/Stop/Reboot/Delete/Replica/Restore/Snapshot); omit the element when no group is set.
  - Resolve the group from state at render time so later `ModifyDBSubnetGroup` updates are reflected.
  - Validate `DBSubnetGroupName` on Create/ReadReplica/Restore (from-snapshot/PIT/S3): unknown -> `DBSubnetGroupNotFoundFault` with rollback; replicas honor an explicit group or inherit the source when omitted; restore paths honor an explicit group, otherwise keep `None`.
  - CloudFormation provisioning reads `DBSubnetGroupName` from `AWS::RDS::DBInstance` properties.
  - Fix workspace tests by setting `db_subnet_group_name` to `None` in `fakecloud-server` test `DbInstance` literals.

<sup>Written for commit 60d407c6062efa2dfb3fd51d8d77a001454ef1eb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/faiscadev/fakecloud/pull/2429?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


vieiralucas:
Thanks a lot for this, @jakshi. Solid, complete fix: storing the subnet group with a serde default, resolving it from state at render time so a later ModifyDBSubnetGroup shows up, rejecting an unknown group with DBSubnetGroupNotFoundFault before any instance is provisioned, and rendering it across every instance response. The tests genuinely fail without the fix. I folded in a couple of follow-ups before merging so the whole surface is consistent: read replicas and the three restore paths (from-snapshot, to-point-in-time, from-s3) now honor an explicit DBSubnetGroupName too (validated the same way, so a restored or replica instance placed in a group reports the right one) rather than inheriting or dropping it, plus the two fakecloud-server test-module DbInstance literals that needed the new field. Merged. Really appreciate the contribution.

jakshi:
Thank you for correcting issues in the code.
I'll try to be more accurate in the future and follow CONTRIBUTING.md more precisely.

PR #2434, opened 2026-07-27
fix(lambda): preserve native Docker host alias

dougludlow:
## Why

Docker Desktop-style runtimes provide `host.docker.internal` natively. Fakecloud's injected `--add-host` mapping could replace that working route, preventing Lambda containers from reaching host services. Image-backed Lambda pulls also need to use the host loopback registry rather than the Lambda callback hostname.

## How

- Preserve an already-resolvable native Docker host alias instead of replacing it.
- Use a dedicated loopback host for local ECR image pulls.
- Permit image-package Lambda functions to invoke without a ZIP deployment package.

## Type

- [x] Bug fix

### DIFF
--- a/crates/fakecloud-core/src/container_net.rs
+++ b/crates/fakecloud-core/src/container_net.rs
@@ -165,7 +165,8 @@ impl HostNetworking {
     /// Resolve networking for `cli`, reading `FAKECLOUD_IN_CONTAINER` from
     /// the process environment.
     pub fn detect(cli: &str) -> Self {
-        let (host_alias, add_host_arg) = resolve_host_alias(cli);
+        let (host_alias, mut add_host_arg) = resolve_host_alias(cli);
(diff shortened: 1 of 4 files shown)


vieiralucas:
Thanks for this, @dougludlow. Real issue: on a Docker-Desktop-class runtime the injected `--add-host host.docker.internal:<vm-bridge-ip>` shadows the alias the daemon already provides natively, breaking Lambda containers' route to host services, and the ECR loopback + image-package invoke fixes are both correct. I pushed one follow-up before merging: the native-alias detection now gates on fakecloud actually running in a container (not just a host-side getaddrinfo), so a bare Linux host with an NXDOMAIN-hijacking resolver can't spuriously drop the bridge mapping native Linux docker needs, and the DNS probe is bounded so a slow resolver can't stall startup. Merged. Really appreciate it.

PR #2436, opened 2026-07-27
fix(cognito): apply pre-token overrides to auth flows

dougludlow:
## Why

Pre-token Lambda overrides were dropped from access tokens issued by SRP, refresh-token, and custom-auth flows. Applications that depend on `cognito:groups` then reject otherwise successful sign-ins.

## How

- Preserve the complete Lambda event response when merging v1/v2 overrides.
- Apply pre-token generation to SRP, custom-auth, and refresh-token token issuance.
- Add regression coverage for v1 group overrides in access tokens.

## Type

- [x] Bug fix

---
## Summary by cubic
Applies Pre Token Generation Lambda overrides across all auth flows (SRP/ADMIN_NO_SRP, custom-auth first-call and multi-round, NEW_PASSWORD_REQUIRED, refresh). Also makes refresh-token rotation atomic to prevent replay and converts remaining paths to async to avoid block_in_place panics.

- **Bug Fixes**
  - Route `AdminInitiateAuth`, `CUSTOM_AUTH` (first-call and multi-round completion), `NEW_PASSWORD_REQUIRED` completion, and `GetTokensFromRefreshToken` through `generate_tokens_with_overrides`.
  - Accept Lambda responses nested under `response` and merge v1/v2 claim overrides for groups and claims.
  - Make refresh-token rotation atomic (old token becomes single-use); add regression tests for admin-initiate/custom-challenge overrides and rotation.

- **Refactors**
  - Convert custom-auth issuance/completion, AdminInitiateAuth, NEW_PASSWORD_REQUIRED, and refresh-token issuance to async and release the state lock before awaiting triggers to avoid `block_in_place` panics (works on current-thread runtimes).

<sup>Written for commit 46dc9df64edb63f164fbf590dc05ffd775e1f84b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/faiscadev/fakecloud/pull/2436?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/fakecloud-cognito/src/service/auth/admin.rs
+++ b/crates/fakecloud-cognito/src/service/auth/admin.rs
@@ -78,7 +78,7 @@ impl CognitoService {
         })
     }
 
-    pub(super) fn admin_auth_verify(
+    pub(super) async fn admin_auth_verify(
         &self,
         input: &AdminAuthInput,
         region: &str,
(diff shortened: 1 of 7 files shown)


vieiralucas:
Thanks for this, @dougludlow. The core catch is right: pre-token Lambda overrides were being dropped from access tokens in the SRP, refresh, and custom-auth flows, and unwrapping the full Lambda event under `response` when merging v1/v2 overrides is the correct fix. Before merging I folded in the remaining token-issuing paths that still bypassed the trigger, so it's now consistent across all of them: custom-challenge completion, NEW_PASSWORD_REQUIRED, AdminInitiateAuth, and GetTokensFromRefreshToken. I also replaced the block_in_place calls with proper async/await (they panic under a current-thread runtime, so the flows are now testable), made the parking_lot guards drop before the trigger await, kept refresh-token rotation atomic with validation so single-use can't be raced, and added flow-level regression tests that decode the access token and assert the overrides land. The hosted-UI OAuth grant endpoints fire the trigger too but need the delivery-context plumbing threaded through, so I left those for a separate change. Merged. Really appreciate the contribution.

PR #2527, opened 2026-09-14
fix(logs): append event segments and reclaim expired logs

Vincent-Wu-Haha:
CloudWatch Logs currently clones and rewrites every retained event after each successful mutation. Sustained ingestion therefore amplifies disk writes with the entire event history, while retention only hides expired events from queries and leaves them in memory and snapshots.

This change appends new events to approximately 4 MiB JSONL segments and atomically commits event byte lengths alongside event-free metadata. Shared save locking covers API calls, EventBridge delivery, CloudFormation hooks, and the retention sweeper. Failed API persistence returns an error instead of acknowledging a successful save.

Retention removes expired events from memory and reclaims fully expired segments, including idle groups on a 60-second sweep. Durable segment cutoffs prevent expired rows from returning after retention is removed or extended, while newly accepted late events remain valid. Sequence watermarks survive pruning. Recovery ignores uncommitted tails and rejects truncated committed data.

Legacy v1/v2 snapshots migrate before the server accepts requests. The old snapshot is removed only after the manifest is durable. This is a one-way format change: users must back up the stopped data directory before upgrading if they need to downgrade. Groups without a retention policy continue to retain events indefinitely; mixed-age segments are reclaimed when all their events expire.

Regression coverage includes incremental append size, unchanged sealed segments, restart ordering, uncommitted tails, truncated data, legacy migration, account isolation, stream recreation, sequence preservation, policy removal, offline expiration, failed persistence, and an official AWS SDK retention/restart test.

Validation:
- `cargo test -p fakecloud-logs -p fakecloud-core -p fakecloud-eventbridge --lib`: 910 passed.
- `cargo test -p fakecloud-e2e --test logs_persistence`: 3 passed.
- `cargo clippy --workspace --all-targets -- -D warnings`: passed.
- `cargo fmt --all --check` and `git diff --check`: passed.
- `cargo build --bin fakecloud`: passed (macOS linker emitted an unwind-table size warning).
- Real-server incremental-write smoke check: after 500 events produced 540,892 bytes of event segments, one more short event appended 86 bytes; all existing event bytes were unchanged.
- `cargo test --workspace`: stops at the existing API Gateway test `data_plane::tests::vpc_link_proxy_resolves_target_nlb_and_attempts_connection` (112 passed, 1 failed in that binary). Its assertion expects a request to localhost port 54321 with no listener to return a connection error. The identical failure reproduces in isolation on unchanged upstream base `1252fc2e7f602c9d1b39a3c35d89fad4872fdf34`, including with `NO_PROXY=*`; the full suite is therefore not green in this environment. No API Gateway code is changed by this PR.


vieiralucas:
Thanks a lot for this, @Vincent-Wu-Haha. Rewriting the whole Logs event history on every mutation was a real scaling problem, and the segment + committed-length manifest design, the resurrection-safe expiration cutoffs, and the thorough recovery tests made it easy to build on.

I pushed a few follow-ups before merging:

- Split each save into a memory-only prepare phase (under the Logs state lock) and a commit phase (segment appends, fsyncs, manifest write, GC) that runs after the lock is released, so disk IO no longer blocks every Logs reader and writer. The store serializes saves with its own lock so a cancelled request cannot let two commits interleave.
- Cached the committed manifest view, so saves no longer re-read the manifest, expiration marks survive a failed commit, and a manifest that landed before a failed directory sync is adopted instead of truncated.
- Skip rewriting an unchanged manifest so the 60s sweep is cheap when idle, while still retrying legacy snapshot removal and segment GC.
- A failed save is logged instead of returning 500, since the mutation was already applied and SDK retries would duplicate events.
- Since retention now deletes data, `retentionInDays` is validated against the AWS value set in PutRetentionPolicy and CloudFormation, and PutLogEvents rejects events older than retention with `expiredLogEventEndIndex` (internal ingestion and EventBridge delivery skip them).
- `storedBytes` accounting is now consistent across ingestion paths and DeleteLogStream, and is normalized once during legacy migration.
- Removed the test-only snapshot-store adapter and pointed the service and EventBridge hook tests at the segmented store.

Merged. Really appreciate the contribution.

PR #2538, opened 2026-09-18
fix(ec2): render iamInstanceProfile on instances, fix IAM profile association state and checks

jakshi:
## Problem

Attach an IAM instance profile to an instance and `describe-instances` still reports `IamInstanceProfile: null`. The association exists (`describe-iam-instance-profile-associations` lists it), but the instance render never emits `<iamInstanceProfile>`. Same for `run-instances --iam-instance-profile`: the launch response omits it because RunInstances renders the instance before writing the association.

Smaller issues in the same family, found while fixing this:

- A direct `associate-iam-instance-profile` stores the association as `associating` and nothing ever advances it. The other two writers store `associated`.
- `replace-iam-instance-profile-association` keeps the old association id and answers `associated`. AWS answers `associating` with a new id (see the doc sample).
- `associate-iam-instance-profile` accepts an unknown instance, and a second profile on an instance that already has one. `replace-iam-instance-profile-association` and `disassociate-iam-instance-profile` accept an unknown association id and answer with a fabricated record.
- Associate and Replace accept a request without `IamInstanceProfile` and store an association with an empty ARN.

## What changed

- `DescribeInstances` resolves the profile from the association map once per request and passes it into the instance render, as it already does for security-group names and block-device mappings. No new field on `Instance`, no snapshot change. Records with state `associating` from older versions still render; they stay untouched, so `describe-iam-instance-profile-associations` keeps reporting them as `associating` until replaced or disassociated.
- `RunInstances` builds the association before rendering, so the launch response carries the element.
- Associate and Replace answer `associating` on the wire and store `associated`, mirroring what Disassociate already does with `disassociating`. Replace retires the old record and mints a new `iip-assoc-` id.
- Associate returns `InvalidInstanceID.NotFound` for an unknown instance and `IncorrectState` for a second profile ("You cannot associate more than one IAM instance profile with an instance", per the AssociateIamInstanceProfile docs). Replace and Disassociate return `InvalidAssociationID.NotFound` for an unknown association id. Associate and Replace return `MissingParameter` when `IamInstanceProfile` is absent (the model marks it required). Error codes are from the EC2 error-code list; the message texts are ours.
- One constructor builds the association record for all four writers (RunInstances, CloudFormation, Associate, Replace), and one helper renders the `<iamInstanceProfile>` fragment for both instance and association responses.

## Test plan

- Unit (handler level, `fakecloud-ec2`): describe shows the profile after associate and drops it after disassociate; `run-instances` response carries it; CloudFormation-provisioned instance shows it; describe shows the new profile after replace; Associate answers `associating` and Describe reads `associated`; Replace answers `associating` with a new id and the old id is gone; Associate on unknown instance -> `InvalidInstanceID.NotFound`; second Associate -> `IncorrectState`; Replace and Disassociate on unknown association id -> `InvalidAssociationID.NotFound`; Associate and Replace without `IamInstanceProfile` -> `MissingParameter`, nothing stored (Replace keeps the old association).
- E2E (`aws-sdk-ec2`, `ec2_instance_control_plane`): run -> associate -> describe asserts the exact `iam_instance_profile().arn()` and `.id()` -> disassociate -> describe asserts `None`.
- Conformance: the Associate, Replace, and Disassociate smoke tests launch an instance (and associate a profile) first instead of using dummy ids, and pass a profile name. Model checksums unchanged.
- With the lookup stubbed to `None`, the describe regression test fails.
- A stored association with state `associating` (written by earlier versions) still renders on the instance.
- `cargo test --workspace --no-fail-fast`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all --check`. Locally, 14801 pass; the only failure is `kinesis_reports_millis_behind_latest_when_limit_truncates`, whose two localhost writes land in the same millisecond. It does not touch EC2. All 94 `fakecloud-tfacc` shards pass locally with terraform 1.16.3.

## Not in this PR

- `iam-instance-profile.arn` / `iam-instance-profile.id` filters on `DescribeInstances`.
- Clearing associations when an instance terminates (already visible today via `describe-iam-instance-profile-associations`).
- Making the rendered profile `id` match the `InstanceProfileId` the IAM service assigned; EC2 mints its own.


vieiralucas:
Thanks a lot for this, @jakshi. The write-up made the review easy: you found the real bug (`describe-instances` reported `IamInstanceProfile: null` forever because the render never emitted the element), and then pulled the whole family of adjacent problems out with it instead of patching just the symptom. Resolving the profile at describe time the way security-group names and block-device mappings already are is the right shape, the `associating`-on-the-wire / `associated`-in-store split matches AWS, and every behavior came with a test.

I pushed a few follow-ups before merging:

- Added `InvalidAssociationID.NotFound`, `IncorrectState` and `InvalidIamInstanceProfileArn.Malformed` to the EC2 shared-error list in the conformance probe. EC2 declares no Smithy errors, so the probe scores any undeclared 4xx as a miss; the new errors would have quietly shed ec2 variants under the flake margin without turning CI red.
- `DescribeIamInstanceProfileAssociations` was ignoring `Filter.N`. The terraform provider reads an instance's profile with `Filter Name=instance-id` and takes item [0], so with two instances it got the wrong association, and with the new Replace semantics that moved one instance's profile onto another. It now honors `instance-id` and `state`, and errors on an unknown explicit id.
- Terminate now drops the association, as AWS does; a stale record kept rendering on the terminated instance and made a later Associate fail forever with the new one-profile check.
- Folded in the two items you listed as not-in-this-PR: the `iam-instance-profile.arn`/`.id` filters, and the terminate cleanup.
- The third one turned out to be reachable too. The rendered profile id is now derived from the profile ARN through a shared `arn::unique_id_for`, so IAM, CloudFormation, EC2 and IMDS all report one `InstanceProfileId` for one profile instead of four different values.
- Validation for the profile reference (empty `Name=` from the SDK builder, a non-instance-profile ARN, a terminated instance), and CloudFormation's update path, which applied `IamInstanceProfile` on create but never on update. That one also needed `ModifyInstanceAttribute` and `CreateTags` added to the EC2 provisioner dispatch, which they had been missing, so any in-place instance stack update was failing outright.

Merged. Really appreciate the contribution, and the level of detail in the PR description.

